PR #226, opened 2014-12-16
discovery of mdns urls

matthewsibigtroth:
Please review @g-ortuno

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
@@ -30,7 +30,6 @@ import android.os.Handler;
 import android.os.ParcelUuid;
 import android.os.Parcelable;
 import android.os.SystemClock;
-import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.View;
@@ -42,7 +41,6 @@ import android.widget.ListView;
 import android.widget.TextView;
 import android.widget.Toast;
 
-import org.uribeacon.beacon.ConfigUriBeacon;
 import org.uribeacon.beacon.UriBeacon;
 import org.uribeacon.scan.compat.ScanRecord;
 import org.uribeacon.scan.compat.ScanResult;


rjsteinert:
Hi all - This sounds very promising. We in the Ground Computing community have been watching this project and this pull request sounds like it may address the issue discussed [here](https://groups.google.com/forum/#!topic/groundcomputing/5MYwcvN8wWY). What is the user story of this feature?

scottjenson:
This may be obvious, but the huge advantage of using BLE is that you don't
need to be on a shared network. We see this as a huge boon to public
devices that have their own connectivity to the internet (we call it
"Kindle connectivity")

However, this has drawbacks in the home where you don't really want your
neighbors to know what devices you have. As the Physical Web is about
putting URLs in the world (and not only about BLE) we have been exploring
other ways to broadcast URLs. MDNS seems like a perfect fit in that it
allows only people authenticated on your wifi to see the URLs being
broadcast on your network.

However, this has immediately exposed a new feature request: to be able to
resolve local URLs. Right now, the client caches and resolves meta data
from each URL by going through a proxy server. This server is on the public
internet so can't see any local URLs.

We've discussed this quite a bit and very much want to support local URLs,
we're just trying to do it properly and would love your feedback. It should
be possible to detect if a URL is definitively local (e.g. IP address
starting with 192, or ending in '.local'). However, the meta data we need
(title, description, favicon) are currently 'scraped' from the target URL
and all of that logic is on the server. We'd rather not have a duplicate of
that code running on the client.

As MDNS allows lots of info to be returned. One solution would be for the
MDNS record to return this info directly in the record. This would add a
small burden to each beacon device but it seems reasonable.

Does this seem practical to you? Is there anything else we should be
looking at?

Scott

scottjenson:
FYI, we're now working on this in 2 stages: the first is to find/resolve
local URLs. This involves detecting that a URL is local and NOT going
through our proxy server. But at the very least, it allows us at least
find/use local URLs.

The next (necessary) stage is to get the meta data from that URL as as well
so it can be displayed properly. That shouldn't be too hard and we'll put
that in shortly afterwards.

Can you explain how this would be useful/impact your project?

Scott

tolson2000:
The other use for the local is for example... My museum would be using our own in house servers for the web pages related to our various exhibits for the most part. Not the public cloud. And I see my airport facilities doing the same thing for many beacons. That is using the facilites local servers for the web pages related to the various customer and activities within the airport. Seems like the ability for URLs to represent in house private IPs is important. And so moving some of the meta data scraping technology into the app for purpose of local URLs is a good idea.

scottjenson:
mDNS is an alternative to BLE. Both cast a URL but BLE has a very limited
payload so ONLY casts the URL, we pickup the meta data from the webpage.

mDNS broadcasts only to your connected wifi and sends out a text record.
We're considering sending out JSON with URL/TITLE/Description/Favicon all
in one blob as we have the room in the record.

tolson2000:
That's kind of what I was thinking. So, using UriBeacons to represent webpages within the facilities can't be dealt with unless our internal servers implement metadata scraping modules themselves.

What does the request packet look like that the smart device sends to your metadata scraping servers? When local or private IP is detected couldn't we send the same request protocol to the local server to see if it would respond with the same return packet you are getting from your servers. Of course, the servers would have to implement similar scraping tech and return return the results using the same protocol.

Does the iOS version of the physical web send the same protocol to Apple owned scraping servers?

PR #259, opened 2015-01-09
slow metadata response fix

matthewsibigtroth:
After scan timeout of 3 seconds, list items that do not have metadata
are shown with just the advertised beacon url. As the metadata is
received (after the 3 seconds), the correlate list item ui is updated
to show that metadata. Also, view recycling is now handled so that list
items are always cleared in getView() before updating (since a recycled
view may contain another url’s metadata from a previous cycle)

Please review @g-ortuno

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
@@ -231,6 +231,7 @@ public class NearbyBeaconsFragment extends ListFragment implements MetadataResol
     mNearbyDeviceAdapter.addItem(url, mockAddress);
     mNearbyDeviceAdapter.updateItem(url, mockAddress, mockRssi, mockTxPower);
     // Inform the list adapter of the new data
+    mNearbyDeviceAdapter.sortDevices();
     mNearbyDeviceAdapter.notifyDataSetChanged();
     // Stop the refresh animation
     mSwipeRefreshWidget.setRefreshing(false);
@@ -464,6 +465,7 @@ public class NearbyBeaconsFragment extends ListFragment implements MetadataResol
     }
 
     public void clear() {
+      mSortedDevices.clear();
       mUrlToDeviceAddress.clear();
       notifyDataSetChanged();
     }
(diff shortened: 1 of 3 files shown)


schilit:
How do you deal with bad urls? Are those handled?

dinhviethoa:
It's handled on the web service side. It will return empty meta data.

matthewsibigtroth:
How are we defining bad urls? Urls that point to nowhere, for example?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -418,29 +411,38 @@ public View getView(int i, View view, ViewGroup viewGroup) {
         view = getActivity().getLayoutInflater().inflate(R.layout.list_item_nearby_beacon, viewGroup, false);
       }
 
+      // Clear the children views content (in case this is a recycled list item view)
+      TextView titleTextView = (TextView) view.findViewById(R.id.title);
+      titleTextView.setText("");
```
g-ortuno: Do you need to set text to reserve the space for the line?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -466,6 +469,10 @@ public void clear() {
       mUrlToDeviceAddress.clear();
       notifyDataSetChanged();
     }
+
+    public void redrawListView() {
```
g-ortuno: Why make a function that just calls one other function?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
matthewsibigtroth: This is to clear the text field, not reserve space.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
matthewsibigtroth: Good question. Mainly for clarity. While really calling notifyDataSetChanged, the reason for it is to redraw the listView. Also, since notifyDataSetChanged is overridden, I can't call the original notifyDataSetChanged (outside the class) without making a new method with a different name.

g-ortuno: 1. By [definition notifyDataSetChanged](http://developer.android.com/reference/android/widget/BaseAdapter.html#notifyDataSetChanged()) implies that the listview should be refreshed. Why not call super.notifyDataSetChanged() directly, instead of calling it through a another function that just calls that one function?

matthewsibigtroth: Oh, because I need to call super.notifyDataSetChanged() from outside the class.

g-ortuno:
LGTM!!

PR #265, opened 2015-01-12
proxy go links

matthewsibigtroth:
Please review @g-ortuno

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
@@ -46,6 +46,8 @@ import org.uribeacon.scan.compat.ScanRecord;
 import org.uribeacon.scan.compat.ScanResult;
 import org.uribeacon.scan.util.RegionResolver;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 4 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -281,12 +283,23 @@ private void openUrlInBrowser(String url) {
...
     }
+    // Route through the proxy server go link
+    url = createUrlProxyGoLink(url);
     // Open the browser and point it to the given url
     Intent intent = new Intent(Intent.ACTION_VIEW);
     intent.setData(Uri.parse(url));
     startActivity(intent);
   }
 
+  private String createUrlProxyGoLink(String url) {
```
dinhviethoa: Is there a way you could use the same method createUrlProxyGoLink() for both usage you have?

matthewsibigtroth: Agreed! I need to figure out a way to call for resources (e.g. R.string.proxy_go_link_base_url)  from a static method (otherwise Android complains). Thanks, I'll look into this.  =)

g-ortuno: What about having the proxy server return the go link in the metadata? That
way we don't repeat code and can control the go link in more interesting
ways.

dinhviethoa:
lgtm.

PR #268, opened 2015-01-14
debug ranging view

matthewsibigtroth:
Please review @g-ortuno


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -153,6 +164,12 @@ public View onCreateView(LayoutInflater layoutInflater, ViewGroup container, Bun
   }
 
   @Override
+  public void onActivityCreated(Bundle savedState) {
+    super.onActivityCreated(savedState);
+    getListView().setOnItemLongClickListener(mAdapterViewItemLongClickListener);
```
g-ortuno: Any reason this is not inside onResume or onCreate?

matthewsibigtroth: onCreate is too early as getListView() throws "java.lang.IllegalStateException: Content view not yet created"
onResume would work, but that would mean rebinding the listener each time which is not necessary

g-ortuno:
LGTM

PR #277, opened 2015-01-27
Proxy server url sorting

matthewsibigtroth:
Please review @g-ortuno

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
@@ -45,8 +45,8 @@ import java.util.Collection;
 
 class MetadataResolver {
   private static final String TAG = "MetadataResolver";
-  //private static final String METADATA_URL = "http://url-caster.appspot.com/resolve-scan";
-  private static final String METADATA_URL = "http://url-caster-dev.appspot.com/resolve-scan";
+  private static final String METADATA_URL = "http://url-caster.appspot.com/resolve-scan";
+  //private static final String METADATA_URL = "http://url-caster-dev.appspot.com/resolve-scan";
   private static final String DEMO_METADATA_URL = "http://url-caster.appspot.com/demo";
   private static RequestQueue mRequestQueue;
   private static boolean mIsInitialized = false;
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
@@ -34,6 +34,8 @@ import org.json.JSONObject;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.util.Collection;
+import java.util.HashMap;
 
 /**
  * Class for resolving url metadata.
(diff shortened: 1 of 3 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
```diff
@@ -90,13 +96,13 @@ public static void findUrlMetadata(final Context context, final MetadataResolver
    *
    * @param url The url for which to request data
    */
-  private static void requestUrlMetadata(String url) {
+  private static void requestUrlMetadata(String url, int txPower, int rssi) {
```
g-ortuno: What about making this three parameters an object. It seems you use all of them together all the time.

matthewsibigtroth: I think an object would be suitable here. Allow me to update. :)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
```diff
@@ -236,6 +249,17 @@ public void onErrorResponse(VolleyError volleyError) {
     );
   }
 
+  public static boolean checkIfMetadataContainsSortingScores(Collection<UrlMetadata> urlMetadataCollection) {
```
g-ortuno: @dinhviethoa this seems hacky and unnecessary. Can you just return whether or not the score is on in the JSON response? I don't see why we have to implement it this way.

dinhviethoa: How about adding a property hasScore to UrlMetadata? That's what iOS implements.

g-ortuno: Since there are two ways of sorting you have to know if the whole list hasScore or not. To do that you still need to check all elements.

dinhviethoa: 1. Checking the whole list is certainly not big trouble since the sorting itself will be slower anyway.

2. We can make assumptions. If we get a score for at least one item. We can assume the feature is enabled on the server and assume that every metadata has a valid score.  

--  
Hoa V. Dinh

g-ortuno: Checking items on the list to make sure a flag is not On doesn't sound like
a good programming practice. Of course you can make assumptions to
implement hacky methods, but that doesn't mean you should. The server
should explicitly communicate whether or not a flag is on.

dinhviethoa: I think that's a fair concern.

Keep in mind that it's a transitional state of the server. Eventually, the server will always send scores. Therefore, I think it would be overkill to create a new webservice to know the server capabilities.

What we can do is making the server transition smooth for the client.

I think we can proceed as follow:
- We can check in that temporary code on the client
- Move the server implementation to proxy sorting
- Then, clean up that temporary code and remove the client-side sorting. We won't have to check whether the score is available or not.

What do you think?

g-ortuno: Ok that sounds reasonable. Then a TODO should be added.

g-ortuno:
lgtm

PR #281, opened 2015-01-29
handling no network

matthewsibigtroth:
If the user taps a url in the list view with no metadata, the browsers
opens to the listed url (instead of the missing siteUrl). The same
metadata check is done for the notifications as well.
Please review @g-ortuno


g-ortuno:
lgtm

PR #290, opened 2015-02-15
Physical Web Chrome App 

louaybassbouss:
This PR contains the initial version on the Physical Web Chrome App. It supports currently only SSDP protocol. Please refer to chrome/README.md and SSDP PR for more details.


scottjenson:
This client is ONLY SSDP so we're reluctant to check it in as it will cause confusion. If it were to also support BLE, we'd accept it.

mmocny:
Scott: How do we feel about just linking to user contributions like this, instead of merging them into this repo?

If we merge them into our repo, we should probably also support them, without understanding exactly how they are implemented.

I think we could create a page which links out to contributor apps like Louay's, and be clear that they aren't vetted.  Then we needn't worry as much about limitations.

scottjenson:
I would agree. I'd be happy to link to this.

louaybassbouss:
Thx @scottjenson @mmocny for your comments. Yes I can put it somewhere on our git and you can link to it. but let me first check if I can add mdns using the [chrome.mdns][1] API. 
Regarding BLE, we need to wait until the required BLE chrome APIs for beacon scan are available.

PS: I see this app is more for experimentation purposes. For best user experience, it makes more sense to implement it as en extension that shows a (physical-web) icon as [Browser Action][2], but the required chrome APIs (UDP, mdns, BLE) are not available for extensions. 

[1]: https://developer.chrome.com/apps/mdns 
[2]: https://developer.chrome.com/extensions/browserAction

Jayseem13:
Well Done!

awennimpagaritse1:
Hello

PR #299, opened 2015-03-20
Make all notifications public on Lollipop

cco3:
I'm aware that it's requested to file an issue first before creating a commit.  This change was mostly just so I could play with the code so I didn't bother creating an issue.

This was tested on a Nexus 5 running Android 5.1

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
@@ -100,6 +100,7 @@ public class UriBeaconDiscoveryService extends Service implements MetadataResolv
   private static final int NON_LOLLIPOP_NOTIFICATION_URL_COLOR = Color.parseColor("#999999");
   private static final int NON_LOLLIPOP_NOTIFICATION_SNIPPET_COLOR = Color.parseColor("#999999");
   private static final int NOTIFICATION_PRIORITY = NotificationCompat.PRIORITY_LOW;
+  private static final int NOTIFICATION_VISIBILITY = NotificationCompat.VISIBILITY_PUBLIC;
   private static final long NOTIFICATION_UPDATE_GATE_DURATION = 1000;
   private boolean mCanUpdateNotifications = false;
   private Handler mHandler;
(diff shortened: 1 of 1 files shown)


g-ortuno:
Could you explain what's the purpose of this change?

cco3:
Sorry, I'll put that in the review message next time.  As the commit message says:
"""
There doesn't seem to be any reason why public beacons would give
anything other than public notifications.  This change makes beacons
publicly visible on the Android 5.0 lock screen.
"""

g-ortuno:
The only time where the contents of the notification will be hidden is if the user explicitly sets "Hide Sensitive Information". If that's the case I don't see why we should force the notification to be public when the user clearly wants privacy.

cco3:
"hide sensitive information" is for hiding things like personal information.  If this wasn't meant to be overridden by notifications that consider themselves public, such functionality wouldn't be allowed by the platform.  If the user *really* wants that level of privacy, that's what the "don't show notifications at all" option is for.

http://developer.android.com/about/versions/android-5.0-changes.html
"If the notification does not contain personal information, or if you want to allow media playback control on the notification, call the setVisibility() method and set the notification's visibility level to VISIBILITY_PUBLIC."

cco3:
An example from the gmail app: with "hide sensitive information" on, single email notifications won't show any content because the subject of the email is personal and sensitive.  However, when you get a notification for multiple emails, the notification shows as it normally would, telling you how many emails you have waiting, because the number of emails isn't that personal.

g-ortuno:
1. But the app could potentially show personal information. The Physical
Web supports other protocols like mDNS where you have to be connected to
the same WiFi network as the device which mean some of the devices are
meant to be private. I can imagine a case where someone has a private
device in their network that wouldn't want to expose in his/her lock
screen.

2. That's a different case. If you want to expose the number of devices
around I have no problem but you are exposing potentially private
information about these devices.

cco3:
OK, if not all beacons are public, that negates the premise of the commit message.

filed #300.

cco3:
OK, taking another shot at this.  I'm leaving all urls discovered through mdns and ssdp as private.

cco3:
But apparently I need an admin to reopen this for me?  Anyway, I've updated the branch.

dinhviethoa:
I'm unable to re-open the pull request as well...

Issue #160, opened 2014-11-01
Make sure URL Shortener network operation is off the ui thread


schilit:
The URL shortener class should not block the main thread. Currently it issues a get() which is blocking. It should probably take an argument of a runnable callback to be invoked by the completion method and use this to call back into the requestor. 

See the documentation around [AsyncTask](http://developer.android.com/reference/android/os/AsyncTask.html) and
the [URL Shortener API](https://developers.google.com/url-shortener/)

@matthewsibigtroth

PR #308, opened 2015-04-01
Do URL shortening/lengthening asynchronously

cco3:
Instead of taking a modified url and then using it to alter the UI,
move that logic out to a callback that is passed to the AsyncTask.

Issue #160

This kind of complicates the logic in BeaconConfigFragment.  We could change things up a bit more by moving the AsyncTask entirely into BeaconConfigFragment, then do all the testing url length etc right in the AsyncTask definition.  UrlShortener would then just be a library of functions without any asynchronous sauce.

Tested on my phone by setting a short url on a beacon and subsequently viewing it.  Not sure what else is needed here.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
@@ -83,6 +83,8 @@ public class BeaconConfigFragment extends Fragment implements TextView.OnEditorA
   private boolean mIsScanRunning;
   private BluetoothAdapter mBluetoothAdapter;
   private Handler mHandler;
+  private UrlShortener.ShortenUrlTask mShortenUrlTask;
+  private UrlShortener.LengthenShortUrlTask mLengthenShortUrlTask;
 
   // Run when the SCAN_TIME_MILLIS has elapsed.
   private Runnable mScanTimeout = new Runnable() {
@@ -166,6 +168,12 @@ public class BeaconConfigFragment extends Fragment implements TextView.OnEditorA
     super.onPause();
     mScanningAnimation.stop();
     scanLeDevice(false);
+    if (mShortenUrlTask != null) {
+      mShortenUrlTask.cancel(true);
+    }
+    if (mLengthenShortUrlTask != null) {
+      mLengthenShortUrlTask.cancel(true);
+    }
     if (mUriBeaconConfig != null) {
       mUriBeaconConfig.closeUriBeacon();
     }
@@ -201,21 +209,38 @@ public class BeaconConfigFragment extends Fragment implements TextView.OnEditorA
     if (status != BluetoothGatt.GATT_SUCCESS) {
       Log.e(TAG, "onUriBeaconRead - error " + status);
     } else {
+      // Create the callback object to set the url
+      UrlShortener.ModifiedUrlCallback urlSetter = new UrlShortener.ModifiedUrlCallback() {
+        public void setUrl(String urlToSet) {
+          // Update the url edit text field with the given url
+          mEditCardUrl.setText(urlToSet);
+          // Show the beacon configuration card
+          showConfigurableBeaconCard();
+        }
+        @Override
+        public void onNewUrl(String newUrl){
+          setUrl(newUrl);
+        }
+        @Override
+        public void onError(String oldUrl) {
+          setUrl(oldUrl);
+        }
+      };
+
       String url = "";
       if (uriBeacon != null) {
         url = uriBeacon.getUriString();
         Log.d(TAG, "onReadUrlComplete" + "  url:  " + url);
         if (UrlShortener.isShortUrl(url)) {
-          url = UrlShortener.lengthenShortUrl(url);
+          mLengthenShortUrlTask = new UrlShortener.LengthenShortUrlTask(urlSetter);
+          mLengthenShortUrlTask.execute(url);
+        } else {
+          urlSetter.onNewUrl(url);
         }
-      }
-      else {
+      } else {
         Toast.makeText(getActivity(), R.string.config_url_read_error, Toast.LENGTH_SHORT).show();
+        urlSetter.onNewUrl(url);
       }
-      // Update the url edit text field with the given url
-      mEditCardUrl.setText(url);
-      // Show the beacon configuration card
-      showConfigurableBeaconCard();
     }
   }
 
@@ -314,6 +339,23 @@ public class BeaconConfigFragment extends Fragment implements TextView.OnEditorA
   }
 
   /**
+   * Write a beacon url to the beacon.
+   */
+  private void setUriBeaconUrl(String url) {
+    try {
+      // Note: setting txPower here only really updates txPower for v1 beacons
+      ConfigUriBeacon configUriBeacon = new ConfigUriBeacon.Builder()
+          .txPowerLevel(TX_POWER_DEFAULT)
+          .uriString(url)
+          .build();
+      mUriBeaconConfig.writeUriBeacon(configUriBeacon);
+    } catch (URISyntaxException e) {
+      e.printStackTrace();
+      Toast.makeText(getActivity(), R.string.config_url_error, Toast.LENGTH_SHORT).show();
+    }
+  }
+
+  /**
    * This is called when the user taps the write-to-beacon button.
    */
   private void saveEditCardUrlToBeacon() {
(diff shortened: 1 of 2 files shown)


dinhviethoa:
That's a great improvement!

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -201,21 +201,29 @@ public void onBeaconConfigReadUrlComplete(ConfigUriBeacon uriBeacon, int status)
     if (status != BluetoothGatt.GATT_SUCCESS) {
       Log.e(TAG, "onUriBeaconRead - error " + status);
     } else {
+      // Create the callback object to set the url
+      UrlShortener.ModifiedUrlCallback urlSetter = new UrlShortener.ModifiedUrlCallback() {
+        public void onNewUrl(String newUrl){
```
dinhviethoa: suggestion: s/onNewUrl/onShortenedUrl/

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -327,21 +335,28 @@ private void saveEditCardUrlToBeacon() {
...
+        try {
+          // Note: setting txPower here only really updates txPower for v1 beacons
+          ConfigUriBeacon configUriBeacon = new ConfigUriBeacon.Builder()
+              .txPowerLevel(TX_POWER_DEFAULT)
+              .uriString(newUrl)
+              .build();
+          mUriBeaconConfig.writeUriBeacon(configUriBeacon);
+        } catch (URISyntaxException e) {
+          e.printStackTrace();
+          Toast.makeText(getActivity(), R.string.config_url_error, Toast.LENGTH_SHORT).show();
```
dinhviethoa: Does anything bad can happen if the user closes the view / activity before the callback is called?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
cco3: But it's not always a shortened url.  This gets used for the url lengthener in the same class.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
cco3: I think because we aren't calling .get() anymore, we don't have to catch any InterruptedException.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
dinhviethoa: ok. sounds good then.

dinhviethoa:
@g-ortuno could you take a look at this one and lgtm/merge if everything looks good? Thanks.

g-ortuno:
Could you test this in a N7 and a N4? We've seen inconsistent behavior in the past and I just want to make sure we aren't breaking other devices again.

g-ortuno:
@cco3 Also slow connections.

dinhviethoa:
Oh. Sure, what happens if there's a network error or the service returns 500?

cco3:
I'm handling service errors better now.  Haven't done the extra tests.

cco3:
I don't know how you compare to my previous commit in github, so I don't know how obvious the change is to you, but see the use of the onError callback.

dinhviethoa:
Usually, the list of commits will appear, unless you rebase, which I think you actually did.
No problem though.

cco3:
Yea, it seemed like this should go in the same commit.

dinhviethoa:
@g-ortuno I'll let you do the final pass on it. It looks good to me for now.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -201,21 +201,35 @@ public void onBeaconConfigReadUrlComplete(ConfigUriBeacon uriBeacon, int status)
...
     } else {
+      // Create the callback object to set the url
+      UrlShortener.ModifiedUrlCallback urlSetter = new UrlShortener.ModifiedUrlCallback() {
+        public void setUrl(String urlToSet) {
+          // Update the url edit text field with the given url
+          mEditCardUrl.setText(urlToSet);
+          // Show the beacon configuration card
+          showConfigurableBeaconCard();
+        }
+        public void onNewUrl(String newUrl){
```
g-ortuno: `@Override` missing.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -201,21 +201,35 @@ public void onBeaconConfigReadUrlComplete(ConfigUriBeacon uriBeacon, int status)
...
+          setUrl(newUrl);
+        }
+        public void onError(String oldUrl) {
```
g-ortuno: `@Override` missing.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -327,21 +341,31 @@ private void saveEditCardUrlToBeacon() {
     if (!URLUtil.isNetworkUrl(url)) {
       url = "http://" + url;
     }
+    // Create the callback object to set the url
+    UrlShortener.ModifiedUrlCallback urlSetter = new UrlShortener.ModifiedUrlCallback() {
+      public void onNewUrl(String newUrl){
```
g-ortuno: `@Override` missing.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -327,21 +341,31 @@ private void saveEditCardUrlToBeacon() {
...
+              .txPowerLevel(TX_POWER_DEFAULT)
+              .uriString(newUrl)
+              .build();
+          mUriBeaconConfig.writeUriBeacon(configUriBeacon);
+        } catch (URISyntaxException e) {
+          e.printStackTrace();
+          Toast.makeText(getActivity(), R.string.config_url_error, Toast.LENGTH_SHORT).show();
+        }
+      }
+      public void onError(String oldUrl) {
```
g-ortuno: `@Override` missing.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
g-ortuno: You probably want to cancel the task if the fragment is closed though.

cco3:
Done

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -327,21 +343,33 @@ private void saveEditCardUrlToBeacon() {
     if (!URLUtil.isNetworkUrl(url)) {
       url = "http://" + url;
     }
+    // Create the callback object to set the url
+    UrlShortener.ModifiedUrlCallback urlSetter = new UrlShortener.ModifiedUrlCallback() {
+      @Override
+      public void onNewUrl(String newUrl){
+        // Write the url to the device
+        try {
```
g-ortuno: You should abstract this and call it `setUriBeaconUrl` and use it here and in line 372.

g-ortuno:
A couple more comments.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
cco3: I'm struggling to find a decent pattern for doing this.  There could be multiple of these that need to be canceled.  I could add them all into an arraylist and iterate through and cancel.  If I wanted to keep that list clean, I would have to check it for finished tasks before adding a new one or I could remove the items from the list when the task completes, but then I'd need to use a synchronized list.  Do you have any suggestions on the best way to go about this?

dinhviethoa: There's probably only one shortener / expander around at a time.

g-ortuno: Why do you say there could be more than one of this tasks running? There are two tasks and each gets executed once: When the Url is read for the beacon and before the Url is written to the beacon.

cco3:
Done...also using Matthew's new error text instead of the vague message I had previously.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -172,6 +174,17 @@ public void onPause() {
   }
 
   @Override
+  public void onDestroy() {
+    super.onDestroy();
+    if (mShortenUrlTask != null) {
```
g-ortuno: I think this one needs to me moved to onPause() since the connection to the beacon gets closed there.

g-ortuno: Once you move it there what happens if you try to cancel an already canceled task? You might need to set `mShortenUrlTask = null` to avoid that.

cco3:
Done.  It just returns false if it's already canceled.
http://developer.android.com/reference/android/os/AsyncTask.html#cancel(boolean)

dinhviethoa:
I'm merging it then!

g-ortuno:
Thanks!

PR #310, opened 2015-04-02
Fixed today view behavior

dinhviethoa:
- Faster update of the beacon list
- Update after the cache has been loaded
review @schilit


g-ortuno:
Why are there so many commits?

dinhviethoa:
I messed up my local copy. I had to rebase... I created an other pull request instead.

PR #312, opened 2015-04-02
Make all UriBeacon notifications public on Lollipop

cco3:
There doesn't seem to be any reason why public beacons would give
anything other than public notifications.  This change makes beacons
publicly visible on the Android 5.0 lock screen when the user has
"hide sensitive notification content" selected.

However, this change keeps urls discovered through mdns or ssdp as
private.

This a continuation of #299 because I can't seem to reopen it.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
@@ -51,6 +51,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
@@ -100,6 +101,7 @@ public class UriBeaconDiscoveryService extends Service implements MetadataResolv
   private static final int NON_LOLLIPOP_NOTIFICATION_URL_COLOR = Color.parseColor("#999999");
   private static final int NON_LOLLIPOP_NOTIFICATION_SNIPPET_COLOR = Color.parseColor("#999999");
   private static final int NOTIFICATION_PRIORITY = NotificationCompat.PRIORITY_LOW;
+  private static final int NOTIFICATION_VISIBILITY = NotificationCompat.VISIBILITY_PUBLIC;
   private static final long NOTIFICATION_UPDATE_GATE_DURATION = 1000;
   private boolean mCanUpdateNotifications = false;
   private Handler mHandler;
@@ -107,6 +109,7 @@ public class UriBeaconDiscoveryService extends Service implements MetadataResolv
   private RegionResolver mRegionResolver;
   private NotificationManagerCompat mNotificationManager;
   private HashMap<String, MetadataResolver.UrlMetadata> mUrlToUrlMetadata;
+  private HashSet<String> mPublicUrls;
   private List<String> mSortedDevices;
   private HashMap<String, String> mDeviceAddressToUrl;
   private MdnsUrlDiscoverer mMdnsUrlDiscoverer;
@@ -195,6 +198,7 @@ public class UriBeaconDiscoveryService extends Service implements MetadataResolv
   private void initializeLists() {
     mRegionResolver = new RegionResolver();
     mUrlToUrlMetadata = new HashMap<>();
+    mPublicUrls = new HashSet<>();
     mSortedDevices = null;
     mDeviceAddressToUrl = new HashMap<>();
   }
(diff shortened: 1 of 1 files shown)


g-ortuno:
@cco3 Could you test this in a N7 and a N4? We've seen inconsistent behavior in the past and I just want to make sure we aren't breaking other devices again.

cco3:
Which N7?  There are 4 of them, haha (grouper, tilapia, flo, deb).  Also, where did the inconsistent behavior lie?

g-ortuno:
Inconsistencies are, sadly, all over the place. In this case I would make sure the notifications are showing up in N7, since that's where we've have problems in the past. Testing on < 21 would also be appreciated.

g-ortuno:
Why do you exactly need an extra parameter? You can just add the url to the `publicUrls` hash set before you call `findUrlMetadata()`.

cco3:
Because that url isn't necessarily the same one that's set later (the one returned by the metadata server).

g-ortuno:
What makes you say that?

cco3:
We send a url to the server in a json object, it gives us a json object back with a url in it.  We don't hang on to the url, so isn't it possible that the service could give us a new URL?  For example, couldn't it give us a lengthened version of a short url?

g-ortuno:
If you step through the code you can see that `onUrlMetadataReceived` is called with the original url as a parameter and the lengthened url is in `urlMetadata.siteUrl`

cco3:
Ok, so the "id" field in the metadata response is guaranteed to be the original url?

g-ortuno:
Yes.
https://github.com/google/physical-web/blob/master/web-service/main.py#L104

I've seen it being use like that throughout the code.

cco3:
Done.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
```diff
@@ -417,6 +422,11 @@ private void updateNearbyBeaconNotification(boolean single, String url, int noti
         .setContentText(description)
         .setPriority(NOTIFICATION_PRIORITY)
         .setContentIntent(pendingIntent);
+    if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+      if (mPublicUrls.contains(url)) {
+        builder.setVisibility(NOTIFICATION_VISIBILITY);
```
g-ortuno: Should be [VISIBILITY_PUBLIC](http://developer.android.com/reference/android/app/Notification.html#VISIBILITY_PUBLIC).

g-ortuno: Please test next time.

g-ortuno:
One comment.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
cco3: Not sure what you mean.  This is VISIBILITY_PUBLIC (See L104)

g-ortuno: You're completely right. Sorry about that. Still I tested it on my Nexus 5 and I don't see any changes in behavior...

cco3: OK, I'll try it again.

cco3: Works right for me.  Is it showing up as content hidden for you?

g-ortuno: Yes. I tried it in two ways:

1. Set `When device is locked` to `Hide sensitive notification content`
2. Turned on `Sensitive` from Apps Notifications Physical Web

In both cases I get `Contents hidden`.

g-ortuno: How many beacons do you have around you?

cco3: I just tried with 1, 2, and many beacons.  Everything looks correct.

I think if you set "sensitive" on the app, then it considers all the notifications sensitive regardless of what the app reports.  Make sure that is turned off and only the lock screen setting itself is set to hide sensitive notification content.

g-ortuno: Weird. Updating the device fixed the issue...

cco3:
Everything works correctly on Nexus4/Nexus7(2013)...note that neither of these have 5.1 yet, so there is no hide sensitive notification content option.

PR #313, opened 2015-04-02
Revert "Ensure the discovery service starts at boot"

g-ortuno:
We need to test this in other devices before merging. I am specially curious as to what happens in N7 2012 since that version doesn't support BLE.


g-ortuno:
@cco3 Could you test in N7 and N4? thanks!

dinhviethoa:
Does this change has anything to do with BLE?

g-ortuno:
Yes. This was only tested on a n5 where BLE is available so the service starts normally. On a N7 2012 there is no BLE so I don't know what happens when this code tries to start a service.

PR #314, opened 2015-04-06
NDEF -> NFC; style nits

sowbug:


scottjenson:
Thanks for this. Much needed and it reads much better.

PR #319, opened 2015-04-10
Server side sorting

mmocny:
Please take a look.  I've not changed much functionally (for now) other than adding server side ranking, but I did restructure to make it a bit easier to track.

I've also added a simple test script to make it easier to iterate.

I've confirmed that this new implementation works with the current Android application with 0 changes to implementation (I just updated the server url to point to my local).

However, the server side ranking does not affect how results are displayed because they are fetched one-at-a-time and sorted on the client.

I'm not 100% convinced that the distance calculation is valid -- but it does sort beacons in the correct order relative to each other, at least when txpower is the same.

### DIFF
--- a/web-service/helpers.py
+++ b/web-service/helpers.py
@@ -32,29 +32,26 @@ def BuildResponse(objects):
     for obj in objects:
         key_id = None
         url = None
-        force = False
+        force = obj.get('force', False)
         valid = True
         siteInfo = None
         rssi = None
         txpower = None
 
-        if "id" in obj:
-            key_id = obj["id"]
-        elif "url" in obj:
-            key_id = obj["url"]
-            url = obj["url"]
+        if 'id' in obj:
+            key_id = obj['id']
+        elif 'url' in obj:
+            key_id = obj['url']
+            url = obj['url']
             parsed_url = urlparse(url)
             if parsed_url.scheme != 'http' and parsed_url.scheme != 'https':
                 valid = False
 
-        if "force" in obj:
-            force = True
-
         # We need to go and fetch.  We probably want to asyncly fetch.
 
         try:
-            rssi = float(obj["rssi"])
-            txpower = float(obj["txpower"])
+            rssi = float(obj['rssi'])
+            txpower = float(obj['txpower'])
         except:
             pass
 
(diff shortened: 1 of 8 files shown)


mmocny:
@dinhviethoa would you mind taking a look?  Theres lots left to do, but hoping for feedback if something is concerning.

The diff vs master is ugly because I moved files around -- however, if you read the patches independently I think its more clean.  I could break the PR up in half to separate the cosmetic move from functional additions if it makes it easier to evaluate.

web-service/helpers.py
```diff
@@ -14,140 +14,119 @@
...
+                device_data["description"] = siteInfo.description
+            if siteInfo.favicon_url is not None:
+                device_data["icon"] = siteInfo.favicon_url
+            if siteInfo.jsonlds is not None:
+                device_data["json-ld"] = json.loads(siteInfo.jsonlds)
+        else:
+            device_data["id"] = url
+            device_data["url"] = url
+        device_data["rssi"] = rssi
+        device_data["txpower"] = txpower
```
dinhviethoa: Could you add a "rank" field to the metadata?
If unknown, we could leave it empty.

mmocny: You mean its position in the sorted list, so that we don't rely on the list ordering directly?

dinhviethoa: I meant a rank value or a score.
Device will be sent in individual requests.
Therefore, at each request, it will sort one device individually.
And there will be no way for the client to sort them.

dinhviethoa:
> jsonlds is not very readable. I can't tell what "lds" is supposed to mean.

It's related to: http://json-ld.org/
Maybe we could add some comment about it.

web-service/helpers.py
mmocny: Not sure how best to do this for long term.  I could just set it to our estimate of distance for now (or normalize that to a less ugly number as you did previous with "Score"), but rank really makes most sense in the context of comparing a set to each other.

I worry that by returning a "score" as opposed to a sorted ordering, we are constraining the future.  However, I understand that this helps the short term, so I will add it.  We can make a different call with v2.

PR #323, opened 2015-04-15
This is a first stab at server side url shortening

mmocny:
The API_KEYS file is required but not included in the repo.

### DIFF
--- a/web-service/app.yaml
+++ b/web-service/app.yaml
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-application: url-caster-dev
+application: url-caster
 version: 1
 runtime: python27
 api_version: 1
--- a/web-service/app.yaml
+++ b/web-service/app.yaml
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-application: url-caster
+application: url-caster-dev
 version: 1
 runtime: python27
 api_version: 1
@@ -24,7 +24,7 @@ handlers:
 - url: /favicon\.ico
   static_files: favicon.ico
   upload: favicon\.ico
-  
+
 - url: /webui
   static_files: static/index.html
   upload: static/index\.html
@@ -32,6 +32,9 @@ handlers:
 - url: /webui.raw
   static_files: static/raw.html
   upload: static/index\.html
+
+- url: /shorten-url
+  script: shortener.app
 
 - url: .*
   script: handlers.app
--- a/web-service/shortener.py
+++ b/web-service/shortener.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+#
+# Copyright 2015 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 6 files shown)


web-service/app.yaml
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-application: url-caster
+application: url-caster-dev
```
dinhviethoa: was that intentional?

web-service/shortener.py
```diff
@@ -0,0 +1,45 @@
...
+                    'Referer': referer
+                })
+        self.response.headers['Content-Type'] = 'application/json'
+        self.response.write(result.content)
+
+################################################################################
+
+app = webapp2.WSGIApplication([
+    ('/shorten-url', ShortURL),
+], debug=True)
```
dinhviethoa: what's debug=True?

web-service/tests.py
```diff
@@ -68,9 +69,17 @@ def testRssiRanking():
         ]
     }
 
-    result = resolveScanForValues(LOCAL_ENDPOINT, values)
+    result = resolveScanForValues(endpoint, values)
     print json.dumps(result, indent=4)
 
+def testUrlShortener(endpoint):
+    values = {"longUrl": "www.github.com/Google/physical-web"}
```
dinhviethoa: s:www.github.com/Google/physical-web:http://www.github.com/Google/physical-web: ?

web-service/shortener.py
mmocny: This I took from the existing handlers.py, but looking at the docs, its:

> The parameter debug=true passed to its constructor tells webapp2 to print stack traces to the browser output if a handler encounters an error or raises an uncaught exception. You may wish to remove this option from the final version of your application.

We should probably turn this off before shipping to prod, but given that its open source I don't think its actually harmful either way.  I suspect we've left this True since the start!

web-service/tests.py
mmocny: Good catch, thanks.

mmocny:
@dinhviethoa Right now the input/output to our server side shortener json is exactly what goo.gl expects and returns.  However, its not exactly ideal.  Should we introduce a custom protocol we control or just leave it?

Also, if we introduce a custom protocol, it will make it possible to swap out the shortening service on the backend without frontend work.  WDYT?

dinhviethoa:
The purpose was to make the change almost seamless on the clients.
I think we could keep a similar input / output even if we use other shortening services.
Maybe we just need to document the input / output?

dinhviethoa:
We can document mandatory fields in the JSON output.

mmocny:
Okay, fair enough.  I'll update the docs and change the name of the
keyfile, providing a sample for the repo.

web-service/tests.py
```diff
@@ -68,9 +73,24 @@ def testRssiRanking():
...
 
-    result = resolveScanForValues(LOCAL_ENDPOINT, values)
+    result = resolveScanForValues(endpoint, values)
     print json.dumps(result, indent=4)
 
+################################################################################
+
+def testUrlShortener(endpoint):
+    values = {
+        'longUrl': 'www.github.com/Google/physical-web'
```
dinhviethoa: still needs http:// or https://

dinhviethoa:
lgtm!

PR #326, opened 2015-04-17
Implemented URL shortener and server-side ranking

dinhviethoa:
- Implemented URL shortener
- server-side ranking
- ability to switch to dev server

review @mmocny @schilit

### DIFF
--- a/ios/PhyWeb/Backend/PWBeacon.m
+++ b/ios/PhyWeb/Backend/PWBeacon.m
@@ -36,9 +36,9 @@
   if (urlString == (NSString *)[NSNull null]) {
     urlString = nil;
   }
-  NSNumber *scoreNumber = info[@"rank"];
-  if (scoreNumber == (NSNumber *)[NSNull null]) {
-    scoreNumber = nil;
+  NSNumber *rankNumber = info[@"rank"];
+  if (rankNumber == (NSNumber *)[NSNull null]) {
+    rankNumber = nil;
   }
 
   if ([title length] == 0) {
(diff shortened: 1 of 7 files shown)


ios/PhyWeb/Backend/PWBeacon.m
```diff
@@ -36,9 +36,9 @@ - (id)initWithUriBeacon:(UBUriBeacon *)beacon info:(NSDictionary *)info {
   if (urlString == (NSString *)[NSNull null]) {
     urlString = nil;
   }
-  NSString *scoreString = info[@"score"];
-  if (scoreString == (NSString *)[NSNull null]) {
-    scoreString = nil;
+  NSNumber *scoreNumber = info[@"rank"];
```
mmocny: Nit: should we still call this a score?

ios/PhyWeb/Backend/PWMetadataRequest.h
```diff
@@ -17,14 +17,15 @@
 #import <Foundation/Foundation.h>
 #import <UriBeacon/UriBeacon.h>
 
-#define METADATA_SERVER_HOSTNAME @"url-caster.appspot.com"
```
mmocny: We should remember to update the README in web-service if this changes names, since we now want to have instructions for how to change the backend.

mmocny:
The sort code is a bit hard to follow, I'd have to dig deeper if you want a toughrough review -- but overall LGTM.

I'll update the backend now so that we can start using this.

PR #335, opened 2015-04-22
Removing the id field from PWS messages

mmocny:
The url field is effectively the id.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
@@ -181,18 +181,14 @@ class MetadataResolver {
 
                   JSONObject jsonUrlMetadata = foundMetaData.getJSONObject(i);
 
+                  String url = jsonUrlMetadata.getString("url");
                   String title = "";
-                  String url = "";
                   String description = "";
                   String iconUrl = "/favicon.ico";
-                  String id = jsonUrlMetadata.getString("id");
                   float score = UNDEFINED_SCORE;
 
                   if (jsonUrlMetadata.has("title")) {
                     title = jsonUrlMetadata.getString("title");
-                  }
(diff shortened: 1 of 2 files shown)


mmocny:
Please review @dinhviethoa @matthewsibigtroth.

I've pushed this version to -dev url-caster so you can test your apps against it easily.  I did not find need to update the iOS app, but have *not* tested the current version (don't have physical device handy).

mmocny:
..actually, just realized that this server update may break existing android apps.  Perhaps we should still send back the id = url for the time being.

dinhviethoa:
Good idea. FYI, I just tested the -dev version with the iOS app. It works well.

mmocny:
Actually, I may now realize what the ID could be used for, although it isn't right now.

The results that come from the metadata server may need a way to associate with the original request.  Right now the request url is not the same as the result url, so there is no way to know the two are related.

See e.g. bug #336.

Perhaps instead of sending & returning an id, we should return "originalUrl" or something?

matthewsibigtroth:
You beat me to it, Michal!
As you mentioned sometimes id and url are not equal in the server response
(e.g. when passing a shortened url to the server, or when the server
resolves to a different endpoint than the passed url). We're keying our
hashmaps by the beacon url on the android side, so for now please keep the
id field. I think your future fix (i.e. "originalUrl") seems very
reasonable.  :)

On Wed, Apr 22, 2015 at 12:05 PM, Hoà V. DINH <notifications@github.com>
 wrote:

> Good idea. FYI, I just tested the -dev version with the iOS app. It works
> well.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/physical-web/pull/335#issuecomment-95306177>.
>
>

mmocny:
Alrighty.  I'm not sure that the server is using id correctly, but I'll update to make sure it does.

mmocny:
..Alrighty I was wrong, its working as intended.  I should fix #336 to leverage this `id` as Android does.

Don't like the name, but we can leave it to not break existing apps.

PR #342, opened 2015-04-29
Fixed nits in mDNS_Support.md

beaufortfrancois:


beaufortfrancois:
Hey @cco3, you're right about the avahi config not being picked up on the fly. 
I've amended my patch to reflect these changes. Please have another look ;)

PR #344, opened 2015-04-30
Support for redirection conditional on rssi and txpower

mmocny:
This PR changes:
- The way we handle uri redirects, simplifying a lot of the code
- Adds two new custom headers: X-PhysicalWeb-Rssi and X-PhysicalWeb-TxPower
- Adds a sample redirection server (local_redirector.py), which redirects only if path_loss <= 50


web-service/helpers.py
```diff
@@ -54,13 +54,10 @@ def append_invalid():
...
 
-        siteInfo = models.SiteInformation.get_by_id(url)
-
-        if force or siteInfo is None:
-            siteInfo = FetchAndStoreUrl(siteInfo, url)
+        siteInfo = GetSiteInfoForUrl(url, rssi, txpower, force_update)
 
         if siteInfo is None:
-            append_invalid()
+            #append_invalid()
```
mmocny: This change is significant.

It means that 404's, 204, etc, are now *filtered out* of the PWS response.  I think this should be the behaviour, but it may affect the experience with e.g. local urls that cannot be hit from PWS but can be from the client on an intranet.

This problem could also be solved locally on the client, perhaps, by detecting local-only urls somehow?

web-service/helpers.py
```diff
@@ -132,42 +130,48 @@ def SortByDistanceCmp(a, b):
 
 ################################################################################
 
-def FetchAndStoreUrl(siteInfo, url):
+def GetSiteInfoForUrl(url, rssi=None, txpower=None, force_update=False):
```
mmocny: This is used to recursively look up in cache after each redirection.  We never cache the redirected value, but we always want to cache the final destination.

web-service/helpers.py
```diff
@@ -132,42 +130,48 @@ def SortByDistanceCmp(a, b):
...
+################################################################################
+
+def FetchAndStoreUrl(siteInfo, url, rssi=None, txpower=None, force_update=False):
     # Index the page
     try:
-        result = urlfetch.fetch(url, validate_certificate = True)
+        headers = {}
+        if rssi is not None and txpower is not None:
+            headers['X-PhysicalWeb-Rssi'] = rssi
+            headers['X-PhysicalWeb-TxPower'] = txpower
```
mmocny: Possibly we also want to add a custom "Content-type" and eventually also "Accept" a custom reply?

web-service/helpers.py
```diff
@@ -132,42 +130,48 @@ def SortByDistanceCmp(a, b):
...
+
+        result = urlfetch.fetch(url,
+                                follow_redirects=False,
```
mmocny: Note, we no longer follow redirect ever.

Previously we would follow redirects to the end, and then make another request to the initial redirector only for a subset of urls.  This way, we don't do any custom handling, nor make needless fetches.

web-service/helpers.py
g-ortuno: Why not add this as a comment?

web-service/helpers.py
```diff
@@ -132,42 +130,48 @@ def SortByDistanceCmp(a, b):
...
-        if real_final_url is None:
-            real_final_url = final_url
-        return StoreUrl(siteInfo, url, final_url, real_final_url, result.content, encoding)
+        assert result.final_url is None
+        # TODO: Use the cache-content headers for storeUrl!
+        return StoreUrl(siteInfo, url, result.content, encoding)
+    elif result.status_code == 204: # No Content
+        # TODO: What do we return?  we want to filter this result out
+        return None
+    elif result.status_code in [301, 302, 303, 307, 308]: # Moved Permanently, Found, See Other, Temporary Redirect, Permanent Redirect
```
mmocny: Unclear to me at this point if we should handle these differently.

Moved permanently usually have a cache-content time of 0, but sometimes not.  We could cache the redirection in that case.

web-service/helpers.py
```diff
@@ -132,42 +130,48 @@ def SortByDistanceCmp(a, b):
...
+        final_url = result.headers['location']
+        # TODO: Most redirects should not be cached, but we should still check!
```
mmocny: Note: Moved permanently usually have a cache-content time of 0, so its rare in practice that these should actually be cached.

web-service/helpers.py
```diff
@@ -226,7 +230,7 @@ def StoreInvalidUrl(siteInfo, url):
 
 ################################################################################
 
-def StoreUrl(siteInfo, url, final_url, real_final_url, content, encoding):
+def StoreUrl(siteInfo, url, content, encoding):
```
mmocny: A note on these values:
`url` was the original uribeacon url
`final_url` was the first redirected value
`real_final_url` was the final redirected value

Now, `url` is just the final redirected value.  We only cache that, no matter which initial uribeacon url it came from.

I'm not 100% sure if its possible for the server to know where it was redirected-from and return a different response based on that.  I looked up `referer` and that is *not* used to give this info.  So I think caching final url regardless of the path it took to get there is fine.

web-service/helpers.py
```diff
@@ -132,42 +130,48 @@ def SortByDistanceCmp(a, b):
 
 ################################################################################
 
-def FetchAndStoreUrl(siteInfo, url):
+def GetSiteInfoForUrl(url, rssi=None, txpower=None, force_update=False):
+    siteInfo = models.SiteInformation.get_by_id(url)
```
dinhviethoa: Is the cache will be empty for most of the request? It will look up using the shortened URL as a key and the metadata are cached using the expanded URL as a key.

web-service/helpers.py
mmocny: Good point!

web-service/helpers.py
mmocny: Exactly.  This is by design, most redirects specifically ask to *not* be cached.

We have TODO's to actually read cache-content headers and respect them, but we aren't doing this even for the final destination, which is really where the win will be.

dinhviethoa: In most of the case, people are forced to use URL shorteners to fit their URLs in the BLE advertising packet and anyway, they would probably like to get fast response for their beacon data.

If you're not associating any cached data to the short URL, in 90% of the case, the cache won't be used, which is questioning the purpose of the cache at all.

mmocny:
Updated to add a redirector right to PWS for testing, making it easier for others to try out.

web-service/helpers.py
mmocny: Not quite.  We still cache the final page content, including title, description etc.  We just don't cache the redirection `Location` http headers.

What will happen is that next time a uribeacon returns e.g. a `goo.gl` link is that we will hit `goo.gl` again (which is by design!), and as long as it returns the same url as it did last time, we will look it up and find it in the cache.

dinhviethoa: Oh sure.

dinhviethoa:
lgtm. You can merge it if you think it's ready.
You can probably add improvement you thought about later.

mmocny:
I think I'll rebase half of this and land it, specifically the changes to caching of redirects etc.

The HTTP headers I think should change to distance & confidence, instead of rssi & txpower, so I won't land that part just yet.

mmocny:
Alrighty, I think this is ready to land.  The latest big changes are:

1. `"rank"` is now a float based on calculated distance, not an int based on path_loss.  The iOS app appears to handle this fine, but I don't have a device to test at the moment.  Will do so Monday morning.
2. `X-PhysicalWeb-Rssi` and `X-PhysicalWeb-TxPower` are replaced with `X-PhysicalWeb-Distance`.  Currently this is calculated on the PWS, but may eventually be passed directly from the client app (which will have the most accurate calculation of distance).
3. experimental features are now disabled unless the application name ends in `-dev`.  This means we can check this in without actually launching features such as the X-PhysicalWeb-Distance to stable.

dinhviethoa:
lgtm.

PR #348, opened 2015-05-05
Switched url-caster to https

beaufortfrancois:
Serving `url-caster.appspot.com` over SSL is cheap and best practise.
Lets' do it.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
@@ -45,9 +45,9 @@ import java.util.Collection;
 
 class MetadataResolver {
   private static final String TAG = "MetadataResolver";
-  private static final String METADATA_URL = "http://url-caster.appspot.com/resolve-scan";
-  //private static final String METADATA_URL = "http://url-caster-dev.appspot.com/resolve-scan";
-  private static final String DEMO_METADATA_URL = "http://url-caster.appspot.com/demo";
+  private static final String METADATA_URL = "https://url-caster.appspot.com/resolve-scan";
+  //private static final String METADATA_URL = "https://url-caster-dev.appspot.com/resolve-scan";
+  private static final String DEMO_METADATA_URL = "https://url-caster.appspot.com/demo";
   private static final int UNDEFINED_SCORE = -1;
   private static RequestQueue mRequestQueue;
   private static boolean mIsInitialized = false;
--- a/android/PhysicalWeb/app/src/main/res/values/strings.xml
+++ b/android/PhysicalWeb/app/src/main/res/values/strings.xml
@@ -39,7 +39,7 @@
     <string name="user_opted_in_flag">userOptedIn</string>
     <string name="physical_web_preference_file_name">physical_web_preferences</string>
     <string name="metadata_loading">loading...</string>
-    <string name="proxy_go_link_base_url">http://url-caster.appspot.com/go?url=</string>
+    <string name="proxy_go_link_base_url">https://url-caster.appspot.com/go?url=</string>
     <string name="ranging_debug_tx_power_prefix">tx: </string>
     <string name="ranging_debug_rssi_prefix">rssi: </string>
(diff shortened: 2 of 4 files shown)


mmocny:
I agree with this change, but differ to @dinhviethoa and @matthewsibigtroth to take at a look at the app changes required.

mmocny:
@dinhviethoa We should investigate this change before end of next week.  If we can make this switch, I think we should try it.

I'm not sure how to do this while transitioning existing clients..  Perhaps adjust clients to use https but leave off the "always" from PWS for now?

mmocny:
`./tests.py -e https://url-caster-dev.appspot.com` works fine, so its a matter of testing all clients.

dinhviethoa:
> @dinhviethoa We should investigate this change before end of next week. If we can make this switch, I think we should try it.

Sounds good. iOS client is already using https.

> I'm not sure how to do this while transitioning existing clients.. Perhaps adjust clients to use https but leave off the "always" from PWS for now?

I also think we should leave off "always" for now.

beaufortfrancois:
For info, `secure: always` is simply a matter of 302 HTTP Redirect when accessing http://url-caster.appspot.com. See https://code.google.com/p/googleappengine/issues/detail?id=2081

I'm not sure therefore why we would want to leave it off.

mmocny:
Thanks for pointing that out.

I think, then, that this is a matter of confirming it will work with all clients.

dinhviethoa:
I don't think POST requests support redirect.  

--  
Hoa V. Dinh

mmocny:
Apparently you can use a 307 redirect (http://programmers.stackexchange.com/questions/99894/why-doesnt-http-have-post-redirect)

beaufortfrancois:
Argh Hoà is right ;)
Should we create a decorator to handle these cases with a 307 redirect or as suggested by both of you, leave "secure: always" off?

mmocny:
I would first create an experiment using 307 and see if our existing clients support it out of the box.

mmocny:
..I can do later today, but if anyone else would like to take a stab, I would be much pleased.

Francois, are you interested?  (I suggest you try this by running a local PWS using, and changing the backend url in ios & android clients.. let me know if you have not done this before).

beaufortfrancois:
Will do that Michal tomorrow morning if that's ok with you.

mmocny:
Thats very appreciated!

beaufortfrancois:
My local PWS now returns 307 HTTP Redirect when a http request comes in. Sadly as you can read at https://android.googlesource.com/platform/frameworks/volley/+/master/src/main/java/com/android/volley/toolbox/BasicNetwork.java#134, Android HTTP Library doesn't know how to handle 307 status codes. That is kinda sad actually...

mmocny:
Sad..

According to [this SO question](http://stackoverflow.com/questions/17481964/android-volley-to-handle-redirect), Volley is using `HttpClient` under the hood, and according to [these docs](http://hc.apache.org/httpcomponents-client-4.2.x/tutorial/html/httpagent.html), it may support `303` code for POST redirects.

Worth a shot?

beaufortfrancois:
Following a 303 status code requires rewriting the method to GET whereas 307 status code keeps the original method ;(

According to http://stackoverflow.com/questions/17481964/android-volley-to-handle-redirect#comment40510412_23140677, *Volley internally uses HttpClient' - that's not accurate, it uses HttpClient only only with API < 9 and uses HTTPUrlConnection otherwise (which got replaced with Okhttp's impl in kit-kat)*
WDYT?

mmocny:
I was going to suggest just switching to use okhttp. Maybe we drop the
redirect for now then..

cco3:
Filed: https://code.google.com/p/android/issues/detail?id=172709

dinhviethoa:
Should we close that pull request then?

mmocny:
Think so, only changes worth making are to `MetadataResolver.java` and `strings.xml`, but that has been refactored and does not apply cleanly.

Leaving it to @cco3 or @beaufortfrancois ?

beaufortfrancois:
@mmocny I've submitted #393 for the Android client.

PR #354, opened 2015-05-05
Add endpoint argument to web-service test

cco3:
The default behavior is to automatically start up a server locally.

### DIFF
--- a/web-service/tests.py
+++ b/web-service/tests.py
@@ -14,15 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import argparse
 import json
+import os
+import subprocess
 import urllib
 import urllib2
 
-################################################################################
 
-LOCAL_ENDPOINT = 'http://localhost:8080'
-REMOTE_ENDPOINT = 'http://url-caster.appspot.com'
-REMOTE_DEV_ENDPOINT = 'http://url-caster-dev.appspot.com'
+LOCAL_TEST_PORT = 9002
 
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


cco3:
@mmocny

web-service/tests.py
```diff
@@ -14,15 +14,15 @@
...
+import subprocess
 import urllib
 import urllib2
 
-################################################################################
 
-LOCAL_ENDPOINT = 'http://localhost:8080'
-REMOTE_ENDPOINT = 'http://url-caster.appspot.com'
-REMOTE_DEV_ENDPOINT = 'http://url-caster-dev.appspot.com'
+LOCAL_TEST_PORT = 9002
```
mmocny: We should update the docs to suggest starting the server with this port.

mmocny: Seems we don't document this at all ([here](https://github.com/google/physical-web/tree/master/web-service#how-to-run-your-own-physical-web-service)).

These are the command I usually use:

To run local development server (from web-service directory):
`dev_appserver.py --host 0.0.0.0 --port 8080 --clear_datastore=yes .`

To deploy dev PWS (from web-service directory):
`appcfg.py --oauth2 --no_cookies -A url-caster-dev update .`

To deploy prod PWS (from web-service directory):
`appcfg.py --oauth2 --no_cookies -A url-caster update .`

web-service/tests.py
```diff
@@ -157,13 +157,66 @@ def testGoLink(endpoint):
...
+
+def main():
+    """The main routine."""
+    # Parse arguments
+    local_url = 'http://localhost:{}'.format(LOCAL_TEST_PORT)
+    parser = argparse.ArgumentParser(description='print things')
+    parser.add_argument(
+            '-e', '--endpoint', dest='endpoint', default='AUTO',
+            help='Which server to test against.\n'
+                 'AUTO:  {} (server starts automatically)\n'
```
mmocny: Oh, I see now.  Cool!

web-service/tests.py
```diff
@@ -157,13 +157,66 @@ def testGoLink(endpoint):
...
-    testDemoData(endpoint)
-    testInvalidData(endpoint)
-    testRssiRanking(endpoint)
-    testUrlWhichRedirects(endpoint)
-    testUrlShortener(endpoint)
-    testRefreshUrl(endpoint)
-    testRedirectWithRssiTxPower(endpoint)
-    testGoLink(endpoint)
+    try:
+        exit(main())
```
mmocny: I've not seen this `exit` convention.. is it just so that code does not execute beyond this?  Is it so that it exists with error code?

mmocny:
Hey, looks really good!  Do you have access to merge yourself?

web-service/tests.py
cco3: Yea, it's my own thing...usually people call exit in the main function, but I prefer to return the exit value from the main function to make the value explicit.  The main difference is that linters can see that you sometimes return an integer and make sure you have returned one by the end of the function.

cco3:
I don't think I do yet.

dinhviethoa:
> I don't think I do yet.

You now have the permissions.

cco3:
Cool, thanks.

PR #355, opened 2015-05-05
Use nose and unittest modules for pws tests

cco3:
Additionally, this change comments out the url shortening test
since that was already broken.

@mmocny

### DIFF
--- a/web-service/tests.py
+++ b/web-service/tests.py
@@ -16,153 +16,207 @@
 #
 import argparse
 import json
+import nose
 import os
 import subprocess
+import unittest
 import urllib
 import urllib2
 
 
 LOCAL_TEST_PORT = 9002
 
-################################################################################
 
-def resolveScanForValues(endpoint, values):
-    req = urllib2.Request(endpoint + '/resolve-scan', json.dumps(values))
-    response = urllib2.urlopen(req)
-    return json.loads(response.read())
+class PwsTest(unittest.TestCase):
+    _HOST = None  # Set in main()
 
-################################################################################
+    @property
+    def HOST(self):
+        PwsTest._HOST
 
-def testDemoData(endpoint):
-    values = {
-        'objects': [
-            { 'url': 'http://www.caltrain.com/schedules/realtime/stations/mountainviewstation-mobile.html' },
-            { 'url': 'http://benfry.com/distellamap/' },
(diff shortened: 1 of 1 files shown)


web-service/tests.py
```diff
@@ -16,153 +16,197 @@
...
+
+class PwsTest(unittest.TestCase):
+    _HOST = None  # Set in main()
+
+    @property
+    def HOST(self):
+        PwsTest._HOST
+
+    def request(self, params=None, payload=None):
+        PATH = getattr(self, 'PATH')
```
mmocny: Why not self.PATH?

cco3: Originally I didn't always have it defined, so there was a default value.  I'll change it.

web-service/tests.py
```diff
@@ -16,153 +16,197 @@
...
+        JSON = getattr(self, 'JSON', False)
+        url = '{}/{}'.format(self.HOST, PATH)
+        if params:
+            url += '?{}'.format(urllib.urlencode(params))
+        args = [url]
+        if payload is not None:
+            args.append(json.dumps(payload))
+        req = urllib2.Request(*args)
```
mmocny: May you add a line to tell that this changes from POST/GET depending on args.

mmocny:
I'm on the run, so haven't run this.  But it looks quite nice, a big improvement.

Curious if the UrlShortener *test* isn't working, or the url shortening on the server is failing?

cco3:
Dunno...it didn't work when I first ran the tests so I didn't really mess with it.  It bombs out when trying to de-serialize the json.  My guess is that you are returning a bare url without quotes or anything.  Let me see if it's that quick of a fix.

mmocny:
Ah yes, we pass through exactly what goo.gl gives.

cco3:
The server returns a 400, so it's not that simple.

mmocny:
Is it possibly that you don't have a config.SECRET.json file?  goo.gl
shortener requires a valid api key.. try running the test against the
deployed url-caster-dev.

cco3:
Ah...that would explain it...yuck...

cco3:
Ok, done.

mmocny:
This is really quite awesome, amazing work.

I keep getting a test failure, but landing this since I think thats a separate issue.

mmocny:
May you also add a document describing how to run tests (including `pip install nose`) some time?

PR #356, opened 2015-05-06
[PWS] Fixed webui form validation

beaufortfrancois:
My previous commit didn't handle correctly webui `form` validation.
This patch does.


mmocny:
I didn't even know you could do this, cool!

PR #357, opened 2015-05-06
[PWS] Handle web manifest name

beaufortfrancois:
Here's a first patch to handle [web manifest](http://www.w3.org/TR/appmanifest/) in PWS.
For now, it only gets `title` based on the web manifest `name`.


mmocny:
Awesome!!

Any ideas about how to get a description?  Firefox and Chrome app manifests support a `"description"` field, but according to the [manifest spec](https://w3c.github.io/manifest/) apparently its not specified for web.

We could check if they key exists, just in case.

Also, perhaps `"short_name"` is a better title and `"name"` could be used as a description?  Not sure..

Issue #220, opened 2014-12-12
Embedding the favicon in the meta data


g-ortuno:
Use "data:" to embed the favicon in the response from the server.

mmocny:
I think we've decided not to pursue this, as per #362?

PR #362, opened 2015-05-07
Switch from favicon URLs to dataURI

beaufortfrancois:
Instead of returning favicon URLs that client needs to fetch, PWS now fetches it first and returns base64 encoded dataURIs in metadata.

BUG=#220

### DIFF
--- a/web-service/helpers.py
+++ b/web-service/helpers.py
@@ -24,6 +24,7 @@ except Exception as e:
         print "Warning: import exception '{0}'".format(e)
 
 from urlparse import urljoin, urlparse
+import base64
 import cgi
 import datetime
 import json
@@ -321,23 +322,26 @@ def StoreUrl(siteInfo, url, content, encoding):
         icon = urljoin(url, '/favicon.ico')
     # make sure the icon exists
     try:
-        result = urlfetch.fetch(icon, method = 'HEAD')
+        result = urlfetch.fetch(icon, validate_certificate=True)
         if result.status_code != 200:
+            icon = None
+        elif 'Content-Type' not in result.headers:
             icon = None
         else:
             contentType = result.headers['Content-Type']
-            if contentType is None:
+            if not contentType.startswith('image/'):
                 icon = None
-            elif not contentType.startswith('image/'):
-                icon = None
-    except:
+            else:
+                icon = 'data:' + contentType + ';base64,' + \
+                       base64.b64encode(result.content)
+    except Exception as e:
         s_url = url
         s_icon = icon
         if s_url is None:
             s_url = '[none]'
         if s_icon is None:
             s_icon = '[none]'
-        logging.warning('icon error with ' + s_url + ' -> ' + s_icon)
+        logging.warning('Icon error with {0}->{1}: {2}'.format(s_url, s_icon, e))
         icon = None
 
     jsonlds = []
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataResolver.java
@@ -18,7 +18,9 @@ package org.physical_web.physicalweb;
 
 import android.content.Context;
 import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
 import android.net.Uri;
+import android.util.Base64;
 import android.util.Log;
 
 import com.android.volley.RequestQueue;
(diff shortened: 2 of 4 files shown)


web-service/helpers.py
```diff
@@ -321,23 +322,24 @@ def StoreUrl(siteInfo, url, content, encoding):
         icon = urljoin(url, '/favicon.ico')
     # make sure the icon exists
     try:
-        result = urlfetch.fetch(icon, method = 'HEAD')
+        result = urlfetch.fetch(icon, validate_certificate=True)
         if result.status_code != 200:
             icon = None
         else:
             contentType = result.headers['Content-Type']
-            if contentType is None:
```
mmocny: Header can still be None?

web-service/helpers.py
```diff
@@ -321,23 +322,24 @@ def StoreUrl(siteInfo, url, content, encoding):
...
+            if not contentType.startswith('image/'):
                 icon = None
-            elif not contentType.startswith('image/'):
-                icon = None
-    except:
+            else:
+                icon = 'data:' + contentType + ';base64,' + \
```
mmocny: Do our existing clients support this format?

web-service/helpers.py
beaufortfrancois: It will trigger an exception but you're right, it's better to be explicit. 
Amended patch to follow.

beaufortfrancois:
Existing clients don't support dataURI. I'm updated those as I speak. Will let you know when CL is updated. 
Thanks for catching that!

g-ortuno:
Yay! Thank you for implanting this!

g-ortuno:
Implementing*

beaufortfrancois:
Hey @g-ortuno, do you mind reviewing Android code at 69e5ef6 before I have a look at the iOS app.
Keep in mind I'm not an Android developer ;)

g-ortuno:
Quick question what does the server return if there is no icon?

beaufortfrancois:
It won't return any 'icon` key when there are none.

g-ortuno:
Mostly nits. Otherwise LGTM

beaufortfrancois:
Nits addressed. Will look at iOS tomorrow.
Thank you Giovanni!

dinhviethoa:
Why don't we just do proxying, it will prevent the size of the metadata to include the size of the image and it will also prevent base64 encoding which is not really efficient.

beaufortfrancois:
I'm not sure proxying is the appropriate solution there.

Let's say we have 10 beacons around us. If we were proxying through http://s2.googleusercontent.com/s2/favicons?domain=http://www.google.com for instance, it would mean 10 additional HTTP requests and their cost associated with this (bandwidth, radio and time).
What do you think?

cco3:
I downloaded 20 various 192x192 pngs, and base64 encoded them to a single json file.  The difference in size between the images together and the json file is 768K vs. 972K.  Seems negligible for such an aggressive test.

I'm in favor of base64 encoding into the metadata.

dinhviethoa:
The main I'm suggesting proxying the images is that we can download metadata first to show something to the user a fast as possible and then, we can download the images later.

cco3:
hm, good point.

beaufortfrancois:
I do agree with you Hoà on the fact it will take longer to get metadata but since the PW App will run in background eventually, speed won't matter that much at this point. 
Avoiding unnecessary HTTP requests seems better in a mobile world.

scottjenson:
The other issue to consider is that some clients (like the current iOS
client) ignores the favicon. In this case, it is forced to download these
images which it just throws away. If the metadata could point to a URL
local to the PWS, it would offer the best of both worlds. Low bandwidth
(for optional clients) and avoiding privacy leakage. It clearly does NOT
help reduce extra http requests (and puts additional load on the server) so
its not clear this is optimal. I just hate to force data on a client that
isn't going to use it.

On Tue, May 12, 2015 at 4:40 AM, François Beaufort <notifications@github.com

beaufortfrancois:
In that case, I would recommend clients could send a "fields" key as part of the HTTP Request POST Data that specifies which fields they want. 
For info, Google APIs handle that very nicely. See https://developers.google.com/apis-explorer/#p/storage/v1/storage.buckets.list

If favicon is not part of "fields", it won't trigger a favicon image fetch.

mmocny:
Good idea.

dinhviethoa:
I think that if we wanted to optimize for the mobile world, we should:
- keep one socket connection opened
- sending first the metadata as a response, it would allow the client show the metadata
- then, sending the favicon: the client would be able to show them as they're streamed
- and probably using protobuf (or ASN1) and not JSON.

That said, what we want to solve here is the privacy issue.
We don't need to solve the bandwidth / latency for now. I think the current state is good enough in regards to bandwidth / latency usage.

And the easiest way I think of without changing the protocol would be to just change the favicon URL in the JSON to http://proxy-server/img?url=http://url-to-the-image

beaufortfrancois:
You've convinced me Hoà. 
I'll close this PR and create another one which will implement favicon proxying.

PR #363, opened 2015-05-07
[PWS] Add checkbox to bypass cache at /webui

beaufortfrancois:
When I'm playing with https://url-caster.appspot.com/webui, I'd love it if it would always fetch website instead of returning stale data. It is particularly true with local servers.

### DIFF
--- a/web-service/static/index.html
+++ b/web-service/static/index.html
@@ -21,6 +21,7 @@ limitations under the License.
     <script>
 
     var resolveScan = function(successCallback) {
+      var forceUpdate = document.getElementById("forceUpdate");
       var resolveQuery = document.getElementById("resolvequery");
       successCallback = successCallback || function() {};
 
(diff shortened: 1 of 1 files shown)


mmocny:
Thats a good point, but then it isn't testing the PWS the same way other clients do.  It could also prevent us seeing bugs.

I usually just use curl if I want to do this:

```
curl --data '{"objects": [{"url":"http://google.ca", "force":true}]}' "http://www.url-caster-dev.appspot.com/resolve-scan"
```

Also, I expect our caching strategy to change over time (like actually respecting cache-control headers)

beaufortfrancois:
Could we add a "force update" checkbox to the UI then?

beaufortfrancois:
What do you think of that? See amended commit.
screenshot 2015-05-07 at 3 27 32 pm

mmocny:
I like it!

PR #366, opened 2015-05-07
[PWS] Check page has some content before parsing

beaufortfrancois:
Go to http://url-caster.appspot.com/webui, enter `http://www.orange.fr` and enjoy seeing the final page being a blank page which used to raise an exception in `GetContentEncoding()`.

Hopefully this patch addresses this issue.


mmocny:
Thanks Francois.  I noticed some 500's only our PWS due to this.

However, while this will prevent the 500, it will still result in non-cached endpoint.  I think this is fine, but may require more investigation..

Also, we should handle more than just empty reply, we need to handle all manner of malformed/missing replies.

PR #367, opened 2015-05-07
Fix url shortening

cco3:
This change also includes a big refactoring commit that turns the MetadataResolver into a generic PwsClient.  The second commit offloads URL shortening to that new generic client.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
@@ -21,6 +21,7 @@ import android.graphics.Bitmap;
 import android.net.Uri;
 import android.util.Log;
 
+import com.android.volley.Request;
 import com.android.volley.RequestQueue;
 import com.android.volley.Response;
 import com.android.volley.VolleyError;
@@ -48,6 +49,7 @@ class PwsClient {
   //private static final String DEV_URL = "http://url-caster-dev.appspot.com";
   private static final String RESOLVE_SCAN_PATH = "resolve-scan";
   private static final String DEMO_RESOLVE_SCAN_PATH = "demo";
+  private static final String SHORTEN_URL_PATH = "shorten-url";
   private static final int UNDEFINED_SCORE = -1;
   private RequestQueue mRequestQueue;
   private Context mContext;
@@ -73,8 +75,12 @@ class PwsClient {
 
   public interface ResolveScanCallback {
     public void onUrlMetadataReceived(String url, UrlMetadata urlMetadata);
-
     public void onUrlMetadataIconReceived();
+  }
+
+  public interface ShortenUrlCallback {
+    public void onUrlShortened(String shortUrl);
+    public void onError(String longUrl);
   }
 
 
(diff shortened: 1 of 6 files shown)


cco3:
Paging @g-ortuno if you have time to take a look

mmocny:
I can take a look in a few hours as well.  Not an expert in our Android app, but should be okay here..

mmocny:
Confirmed this is working well, woot!

android/PhysicalWeb/app/src/main/AndroidManifest.xml
```diff
@@ -16,6 +16,7 @@
 
     <application
         android:allowBackup="true"
+        android:name=".PhysicalWeb"
```
mmocny: Is the `.` intentional?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
```diff
@@ -37,29 +37,33 @@
...
+ * For the largest use, metadata requests, the physical web service
+ * scrapes the page at the given url for its metadata.
  */
-
-class MetadataResolver {
-  private static final String TAG = "MetadataResolver";
-  private static final String METADATA_URL = "http://url-caster.appspot.com/resolve-scan";
-  //private static final String METADATA_URL = "http://url-caster-dev.appspot.com/resolve-scan";
-  private static final String DEMO_METADATA_URL = "http://url-caster.appspot.com/demo";
+class PwsClient {
```
mmocny: May we scan the codebase for references to this old class.. I think it's used in some docs.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
```diff
@@ -37,29 +37,33 @@
...
+  private static final String TAG = "PwsClient";
+  private static final String PROD_URL = "http://url-caster.appspot.com";
+  //private static final String DEV_URL = "http://url-caster-dev.appspot.com";
```
mmocny: Drive-by future feature request: iOS switches to DEV backend when in dev mode.. can we implement something like this for Android, so you can test PWS features without an App re-build?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
```diff
@@ -37,29 +37,33 @@
...
+  private static final String RESOLVE_SCAN_PATH = "resolve-scan";
+  private static final String DEMO_RESOLVE_SCAN_PATH = "demo";
+  private static final String SHORTEN_URL_PATH = "shorten-url";
   private static final int UNDEFINED_SCORE = -1;
```
mmocny: Drive-by bug report: We don't use score, we use "rank".

android/PhysicalWeb/app/src/main/AndroidManifest.xml
cco3: Yea, see .MainActivity on L24

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
cco3: dev mode?  You mean like when you have developer options enabled on android?  I have no idea if you can even see that, but I'd be willing to look into it.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
cco3: Done

g-ortuno:
There is a bug similar to the bug you had last time. If you get the shortened url after the activity gets destroyed the app crashes. You can test it with a slow network:

1. Open config mode
2. Enter long url
3. Save long url
4. Press back before the URL gets shortened.
5. App crashes

Crash log:
```java
java.lang.NullPointerException: Attempt to invoke virtual method 'void org.uribeacon.config.GattRequestQueue.add(android.bluetooth.BluetoothGatt, org.uribeacon.config.GattRequestQueue$RequestType, android.bluetooth.BluetoothGattCharacteristic)' on a null object reference
            at org.uribeacon.config.GattService.writeCharacteristic(GattService.java:85)
            at org.uribeacon.config.ProtocolV1.writeUriBeacon(ProtocolV1.java:92)
            at org.uribeacon.config.UriBeaconConfig.writeUriBeacon(UriBeaconConfig.java:97)
            at org.physical_web.physicalweb.BeaconConfigFragment.setUriBeaconUrl(BeaconConfigFragment.java:347)
            at org.physical_web.physicalweb.BeaconConfigFragment.access$900(BeaconConfigFragment.java:67)
            at org.physical_web.physicalweb.BeaconConfigFragment$5.onUrlShortened(BeaconConfigFragment.java:372)
            at org.physical_web.physicalweb.PwsClient$1.onResponse(PwsClient.java:119)
            at org.physical_web.physicalweb.PwsClient$1.onResponse(PwsClient.java:108)
            at com.android.volley.toolbox.JsonRequest.deliverResponse(JsonRequest.java:65)
            at com.android.volley.ExecutorDelivery$ResponseDeliveryRunnable.run(ExecutorDelivery.java:99)
```

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
mmocny: I meant an application-level dev mode.  I haven't seen it, but I think there is an option from the dropdown on iOS to enter dev/debug mode or something

g-ortuno: I know @matthewsibigtroth added a dev mode as well. If you long press the url list in android it goes into dev mode.

cco3: Oh, OK.

mmocny: @g-ortuno Oh, sweet!

cco3:
Fixed using volley's tag-based cancellation.

g-ortuno:
Any reason why are you using this instead of the singleton pattern? It seems the singleton pattern is the recommended way:

https://developer.android.com/training/volley/requestqueue.html#singleton

cco3:
Yes, I made PwsClient a singleton.

cco3:
But if the requestqueue is in a singleton, that means it can't know what activities it gets requests from...that's why you use tag-based cancellation.

g-ortuno:
I mean why are you subclassing Application, which seems to be discouraged, and not using a static singleton class?

https://developer.android.com/training/volley/requestqueue.html#singleton

cco3:
I don't create the RequestQueue in the subclassed Application.  I'm assuming that's discouraged because you shouldn't create it until it's needed.

cco3:
Oh, I see in the link to the Application documentation that they prefer you pass along context...

OK, I'll do it that way, but that seems like a pain.

g-ortuno:
I don't really have a preference. I was just curious why you picked one over the other. If you think that the current pattern is better suited for this then LGTM

cco3:
OK, fixed it up.  Tested, everything works.

g-ortuno:
Feel free to merge once travis finishes.

PR #368, opened 2015-05-07
Copy Edit README.md

jpmedley:


scottjenson:
Looks good. Thanks @jpmedley! It's clear that URLs vs URL's is a heated web topic but as smarter people than me say the URL's is correct, I'm not about to rock the boat.

PR #371, opened 2015-05-08
Removing *.iml files and adding to gitignore

mmocny:


mmocny:
I'm not 100% sure about this one.. but it appears Android Studio is auto-generating these for me, and if we don't remove them, I get codebase changes every time I try to build the project.

Some quick research suggests it's advised to `.gitignore` these files, but I'm not an expert.  It works for me, but looking to confirm others can still import this project after this change.

g-ortuno:
The documentation says to include them and when you create a new project
the generated .gitignore doesn't include .iml files.

https://intellij-support.jetbrains.com/entries/23393067

mmocny:
Thats the case from 2013, but according to this [Stack Overflow question](http://stackoverflow.com/questions/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project), the top answer includes it in gitignore, and the most recent comment also suggests users are adding it to gitignore in practice while sharing with team, and without trouble.

Do we understand the file to know if it is useful in any way?

mmocny:
Apparently those files include info like android sdk version and certain file paths.. so deleting them may not make sense if we use non-default values.  I'll close this PR.

However, those file are also not supposed to change, so something funky is happening on my machine.  Perhaps I'm using a newer version of Andrew Studio which is adding new tags which should just be committed once.. I'll create a new PR for that.

g-ortuno:
Yeah. Every time I update android studio I have to make a pull request to
all the repos I'm working on to change the iml files.

devunwired:
These files are generated from the `build.gradle` by Android Studio, which is a project's source of truth.

Remember that, while Studio is based on IntelliJ IDEA, not all the same rules apply to both. Studio wouldn't actually need IML files at all if the IntelliJ platform didn't rely on them for certain dependency mappings.

PR #372, opened 2015-05-08
IML file updates

mmocny:
This is in response to discussion in #371.


mmocny:
The changes here are magic done by latest version of Android Studio.

Issue #365, opened 2015-05-07
PWS tests.py fails on `test_redirect_with_rssi_tx_power` when auto-starting server 


mmocny:
The `test_redirect_with_rssi_tx_power` test is supposed to only work with `app.yaml` set to `-dev` (because it hits the `experimental/` endpoint of the PWS), but it isn't working with the server auto-starts (by running `./tests.py`), even when I change the `app.yaml`.

If I start a `dev_appserver.py` manually and run `tests.py -e LOCAL` it works fine!

mmocny:
Ideally we can skip all the `-dev` tests if app.yaml is not set for it.

cco3:
> with app.yaml set to -dev

What field is that?

mmocny:
App name.  You can check helpers.py if I recall correct where it checks for
"experimental" mode.

cco3:
Skip them because that functionality isn't yet in prod?  Or does the distinction go beyond that?

cco3:
Also, do you have a quick fix for: "OperationalError: database is locked"

mmocny:
I've not seen that error before, sorry..

mmocny:
Yeah just skip them. Would be nice if test suite could display they were
skipped, but it's fine either way.

PR #378, opened 2015-05-12
[PWS] Start/stop auto test server more reliably

cco3:
This makes the tests a little more reliable, but I'm still not sure how to reliably reproduce #365

### DIFF
--- a/web-service/tests.py
+++ b/web-service/tests.py
@@ -18,7 +18,9 @@ import argparse
 import json
 import nose
 import os
+import signal
 import subprocess
+import sys
 import unittest
 import urllib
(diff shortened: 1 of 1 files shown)


mmocny:
When I try to run this, I always get "Rogue server is already running", even though I can bind port fine without this patch.

cco3:
ps -ef | grep dev_appserver

Is there anything there?

mmocny:
There was, killing it resolves the issue, thanks.

Curious though why it would ignore the bind issue in the other case?

cco3:
Yea...that confuses me...I didn't change the port number.

PR #380, opened 2015-05-12
[PWS] Added Physical Web User Agent

beaufortfrancois:
Some websites rely on the "Mozilla5/0" part of the User-Agent string to serve web content. Let's add a Physical Web User Agent to fix that issue.

Moreover, this could lead websites in general to serve specific content (small and fast) for those who examine HTTP Headers User-Agent. What do you think?

### DIFF
--- a/web-service/helpers.py
+++ b/web-service/helpers.py
@@ -33,6 +33,7 @@ import lxml.etree
 ################################################################################
 
 ENABLE_EXPERIMENTAL = app_identity.get_application_id().endswith('-dev')
+PHYSICAL_WEB_USER_AGENT = 'Mozilla/5.0 (compatible; PhysicalWeb/1.0; +http://physical-web.org)'
 
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


cco3:
LGTM

beaufortfrancois:
Can we update url-caster.appspot.com and url-caster-dev.appspot.com as well with this change?

mmocny:
Guys I'm not sold on this change.

Specifically, there will be many user agents acting as the physical web metadata resolver, and I'm not sure we should be making the promise that this will be the user agent to web servers.

Imagine someone wants to make an app that resolves from the phone directly, doesn't want to use a middleman -- should they get different results?  Or a company wants to make a private proxy for their employees, or whatever..

mmocny:
..I think instead we may in the future  support a custom data type (perhaps), and signal that fact by passing the `Accept` header, so that the web server could return a custom manifest or something.

Speculative at this point, but just throwing this out there.

dinhviethoa:
@mmocny feel free to revert if you're not convinced by this change.

mmocny:
Reverted in 293a902536e9f7a8038232396d0c06a960fa0f45.

Sorry I just want to consider the implications before committing to this.

beaufortfrancois:
How about we remove the `(compatible; PhysicalWeb/1.0; +http://physical-web.org)` part of the Physical Web Service User Agent and simply leave `Mozilla/5.0` part so that servers that rely on it to show web content to their users could still be parsed. An example of this is french ISP: http://www.orange.fr

mmocny:
Sure, thats a fair initial value, we can iterate from there.

beaufortfrancois:
Here's a PR which adds only the `Mozilla/5.0` User Agent: https://github.com/google/physical-web/pull/390/files

Issue #305, opened 2015-03-26
Favicon through proxy


cco3:
All the needed metadata of a site is grabbed through a proxy, but then (at least on the android app), the favicon is fetched directly from the app.  This seems to negate the point of using a proxy in the first place.  It seems like the proxy should actually give us the favicon instead of just the url to the favicon.

scottjenson:
Agreed, this shouldn't happen. There are two paths to fix this, a) make sure the server gets it and b) give the server more options to find it robustly (see #301) by using manifest. Both are being filed as bugs/feature requests, thank you!

g-ortuno:
Yes!! I even opened a bug #220 about a possible implementation...

mmocny:
Fixed!

PR #383, opened 2015-05-13
[PWS] Added favicon proxy handler

beaufortfrancois:
As suggested by Hoà https://github.com/google/physical-web/pull/362#issuecomment-101376415, here's a way to get a uribeacon favicon url through PWS.

BUG=#305

### DIFF
--- a/web-service/helpers.py
+++ b/web-service/helpers.py
@@ -23,6 +23,7 @@ except Exception as e:
     else:
         print "Warning: import exception '{0}'".format(e)
 
+from urllib import quote_plus
 from urlparse import urljoin, urlparse
 import cgi
 import datetime
@@ -83,7 +84,7 @@ def BuildResponse(objects):
         if siteInfo.description is not None:
             device_data['description'] = siteInfo.description
         if siteInfo.favicon_url is not None:
-            device_data['icon'] = urljoin(BASE_URL, '/favicon?url=' + siteInfo.favicon_url)
+            device_data['icon'] = urljoin(BASE_URL, '/favicon?url=' + quote_plus(siteInfo.favicon_url))
         if siteInfo.jsonlds is not None:
             device_data['json-ld'] = json.loads(siteInfo.jsonlds)
         device_data['distance'] = distance
--- a/web-service/helpers.py
+++ b/web-service/helpers.py
@@ -34,6 +34,7 @@ import lxml.etree
 
 ENABLE_EXPERIMENTAL = app_identity.get_application_id().endswith('-dev')
 PHYSICAL_WEB_USER_AGENT = 'Mozilla/5.0 (compatible; PhysicalWeb/1.0; +http://physical-web.org)'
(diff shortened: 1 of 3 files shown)


web-service/helpers.py
```diff
@@ -37,9 +37,14 @@
...
 ################################################################################
 
-def BuildResponse(objects):
+def BuildResponse(objects, request_headers):
     metadata_output = []
 
+    if 'Origin' in request_headers:
+        base_url = request_headers['Origin']
+    else:
+        base_url = 'https://' + app_identity.get_application_id() + '.appspot.com'
```
mmocny: Nit: can we make this a constant up top?  You have a separate FR to change the base_url, so it will be good to have only place to adjust this

web-service/helpers.py
```diff
@@ -37,9 +37,14 @@
 
 ################################################################################
 
-def BuildResponse(objects):
+def BuildResponse(objects, request_headers):
     metadata_output = []
 
+    if 'Origin' in request_headers:
+        base_url = request_headers['Origin']
```
mmocny: Are we sure we prefer to use Origin header?  I thought that is set to the origin of the request, which would not likely have a `/favicon?` endpoint..

web-service/helpers.py
```diff
@@ -380,6 +385,18 @@ def StoreUrl(siteInfo, url, content, encoding):
 
 ################################################################################
 
+def FaviconUrl(url):
+    # Make sure we only fetch favicons we've already added to our database.
```
mmocny: Nit: "..fetch favicons *for sites* we've already.."

mmocny: Also.. is there any reason why this is important to constrain?

If someone wants to abuse the feature, all they have to do is call "resolve-scan" once for a given url.  Perhaps we need to rethink this approach..

beaufortfrancois: This check is only made because the `favicon_url` is added into the database only if it's a valid favicon image based on the only facts its is indeed a publicly accessible URL that returns a blob whose contentType starts with `image/`.

beaufortfrancois:
I've addressed your feedback and added one answer to the constrain comment. Please let me know.

web-service/helpers.py
```diff
@@ -82,7 +83,7 @@ def append_invalid():
         if siteInfo.description is not None:
             device_data['description'] = siteInfo.description
         if siteInfo.favicon_url is not None:
-            device_data['icon'] = siteInfo.favicon_url
+            device_data['icon'] = urljoin(BASE_URL, '/favicon?url=' + siteInfo.favicon_url)
```
dinhviethoa: siteInfo.favicon_url probably needs to be URL encoded.

dinhviethoa:
@mmocny do you have additional comments?
Also, I think the patch needs to be rebased: It's showing "We can’t automatically merge this pull request.".

beaufortfrancois:
I've added URL encoding for favicon_url and rebased it.

PR #384, opened 2015-05-13
Fix android crash by ensuring Nsd listener liveness when stopping discovery

beaufortfrancois:
Calling `mNsdManager.stopServiceDiscovery(null);` raises an exception and crashes the android app. Let's check first `mDiscoveryListener` is not null.


dinhviethoa:
@beaufortfrancois could you take a look at @cco3 questions?

beaufortfrancois:
@cco3 I've been trying to reproduce the Android Crash today and I've failed. Somehow, `mDiscoveryListener` was null last week when I was playing with the Android App and this fix made it go away. However, today, I'm unable to replicate this crash. Therefore, I'll close this PR and resubmit when/if I can share with you what was happening behind the scenes.

Regarding the SO bug, it could be related to it but I'm not sure there. Sorry for not helping there ;(

cco3:
No worries, thanks for your efforts.

PR #385, opened 2015-05-13
Fix #365: Don't run experimental tests when server not in dev mode

### LINKED ISSUE #365: PWS tests.py fails on `test_redirect_with_rssi_tx_power` when auto-starting server

mmocny:
We may want to clean up tests if ever there are more experimental-only features.  But I think there is nothing to warrent that just yet.

### DIFF
--- a/web-service/README.md
+++ b/web-service/README.md
@@ -90,6 +90,5 @@ Physical Web Service, you can change the endpoint by modifying:
 1. Make sure you have created config.SECRET.json from the config.SAMPLE.json
    file.
 2. Install nose `pip install nose`
-2. Install pyyaml `pip install pyyaml`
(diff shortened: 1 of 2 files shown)


mmocny:
@cco3

dinhviethoa:
any update on this one?

mmocny:
Thanks for poke, time to land.

PR #386, opened 2015-05-14
This PR adds support for "unresolved" return key from resolve-url

mmocny:
For urls which we could not look up (i.e. 404 errors), we add them
to the list of unresolved.

This is distinct from urls which we could fetch, but filtered out of
results (i.e. 500 errors, 204 No Data)..

### DIFF
--- a/web-service/handlers.py
+++ b/web-service/handlers.py
@@ -62,10 +62,8 @@ class ResolveScan(webapp2.RequestHandler):
         except:
             objects = []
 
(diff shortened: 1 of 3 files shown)


mmocny:
@dinhviethoa 

I haven't tried all the clients yet, will take a stab at that tomorrow morning.

dinhviethoa:
iOS behaves correctly with this change. lgtm.

mmocny:
Thanks Hoa

PR #387, opened 2015-05-14
If a beacon has no rank, we set a rank of 1000 for it

dinhviethoa:
review: @mmocny

### DIFF
--- a/ios/PhyWeb/UI/PWBeaconsViewController.m
+++ b/ios/PhyWeb/UI/PWBeaconsViewController.m
@@ -327,92 +327,28 @@
 
 // Sort results by RSSI value.
 - (void)_sort {
-  BOOL stableMode = [[PWBeaconManager sharedManager] isStableMode];
-
-  BOOL hasRank = NO;
   for(PWBeacon * beacon in _beacons) {
-    if ([beacon hasRank]) {
-      hasRank = YES;
-      break;
-    }
-  }
-  if (hasRank) {
-    for(PWBeacon * beacon in _beacons) {
-      if (![beacon hasRank]) {
-        [beacon setHasRank:YES];
-        [beacon setRank:1000];
-      }
+    if (![beacon hasRank]) {
+      [beacon setHasRank:YES];
+      [beacon setRank:1000];
     }
   }
   [_beacons sortUsingComparator:^NSComparisonResult(id obj1, id obj2) {
     PWBeacon *beacon1 = obj1;
     PWBeacon *beacon2 = obj2;
-    if (stableMode) {
-      if ([beacon1 hasRank]) {
-        PWBeacon *beacon1 = obj1;
-        PWBeacon *beacon2 = obj2;
-        double diff = [beacon1 rank] - [beacon2 rank];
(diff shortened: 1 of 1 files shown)


ios/PhyWeb/UI/PWBeaconsViewController.m
```diff
@@ -329,6 +329,21 @@ - (void)_updateBeaconsNow {
 - (void)_sort {
   BOOL stableMode = [[PWBeaconManager sharedManager] isStableMode];
 
+  BOOL hasRank = NO;
+  for(PWBeacon * beacon in _beacons) {
+    if ([beacon hasRank]) {
+      hasRank = YES;
```
mmocny: Is this basically checking if the PWS support ranking?

ios/PhyWeb/UI/PWBeaconsViewController.m
```diff
@@ -329,6 +329,21 @@ - (void)_updateBeaconsNow {
...
+      break;
+    }
+  }
+  if (hasRank) {
+    for(PWBeacon * beacon in _beacons) {
+      if (![beacon hasRank]) {
+        [beacon setHasRank:YES];
```
mmocny: When would server not have rank?

dinhviethoa: I think it happens when the beacon is unresolved.

g-ortuno: We do the same in the Android client: https://github.com/google/physical-web/pull/277/files

From what I understood it was supposed to be used while we transitioned to server side ranking.

ios/PhyWeb/UI/PWBeaconsViewController.m
dinhviethoa: It's when PWS doesn't returns the rank consistently (for example, on failure).

mmocny: Not sure I understand, but you're the boss ;)

Patch looks fine, pull it.

dinhviethoa: I think that with the latest change on the metadata server, we're probably not seeing the issue any more.
What happened before was append_invalid() added beacons that did not have a rank.
(see https://github.com/google/physical-web/commit/62321566ee833815864bf95bbc63adc65e1f1868#diff-32f4822b465eace8648030b1fec3eb2c)

I'm not sure if we should land it since it's going to be improved on server side very soon.

mmocny: Fair -- though you still want to set rank to 1000 for those, no?  So we could just say "if no rank then 1000" for all values without checking for server side support?

dinhviethoa: We can start assuming that PWS will return a rank.
And if beacon have no rank assign them 1000.
I'll change this pull request.

dinhviethoa:
Updated.

mmocny:
Wow so much cleaner!

PR #388, opened 2015-05-15
Add new displayUrl value to resolve-scan

mmocny:


mmocny:
@dinhviethoa @cco3 

I added a new "displayUrl" value which clients should be changed to use to show in the ui.

Basically:  `<a href={{url}}>{{displayUrl}}</a>`

For now, both values are equal, since we do not want to change existing clients.  We can start to make the shift eventually.  It won't affect correctness, but old clients may start seeing shortened urls.  Not a major concern.

dinhviethoa:
Just to let you know. I re-discovered last week that the iOS client was expanding the URL broadcasted by the beacon. Therefore, it probably won't affect the iOS client.
Not sure about the Android one.

mmocny:
@dinhviethoa So I understand, you mean that we fetch HEAD of each url that is returned to see if they redirect, and resolve all the way down?

I don't think this is desireable, but we should factor that in for old clients.

cco3:
We just use what pws sends down in android.

dinhviethoa:
> I don't think this is desireable, but we should factor that in for old clients.

I agree we should fix that.

dinhviethoa:
any update on this one?

PR #382, opened 2015-05-12
[PWS] Fixed encoding on title and description metadata

beaufortfrancois:
Go to http://url-caster-fr.appspot.com/webui and enter "http://www.blackanddecker.fr/".
Before:
```
{
  "metadata": [
    {
      "description": "BLACK+DECKERâ¢ est le premier fournisseur mondial d'outils Ã©lectriques et d'accessoires.",
      "url": "http://www.blackanddecker.fr/",
      "rank": 1000,
      "icon": "http://www.blackanddecker.fr/favicon.ico",
      "title": "Outils Ã©lectriques & Outils de jardin â BLACK+DECKERâ¢",
      "id": "http://www.blackanddecker.fr/"
    }
  ]
}
```
After:
```
{
  "metadata": [
    {
      "description": "BLACK+DECKER™ est le premier fournisseur mondial d'outils électriques et d'accessoires.",
      "url": "http://www.blackanddecker.fr/",
      "rank": 1000,
      "icon": "http://www.blackanddecker.fr/favicon.ico",
      "title": "Outils électriques & Outils de jardin – BLACK+DECKER™",
      "id": "http://www.blackanddecker.fr/"
    }
  ]
}
```

PR #389, opened 2015-05-18
[PWS] Fixed encoding on title and description metadata [Take 2]

beaufortfrancois:
Go to http://url-caster-fr.appspot.com/webui and enter "http://www.blackanddecker.fr/".
Before:
```
{
  "metadata": [
    {
      "description": "BLACK+DECKERâ¢ est le premier fournisseur mondial d'outils Ã©lectriques et d'accessoires.",
      "url": "http://www.blackanddecker.fr/",
      "rank": 1000,
      "icon": "http://www.blackanddecker.fr/favicon.ico",
      "title": "Outils Ã©lectriques & Outils de jardin â BLACK+DECKERâ¢",
      "id": "http://www.blackanddecker.fr/"
    }
  ]
}
```
After:
```
{
  "metadata": [
    {
      "description": "BLACK+DECKER™ est le premier fournisseur mondial d'outils électriques et d'accessoires.",
      "url": "http://www.blackanddecker.fr/",
      "rank": 1000,
      "icon": "http://www.blackanddecker.fr/favicon.ico",
      "title": "Outils électriques & Outils de jardin – BLACK+DECKER™",
      "id": "http://www.blackanddecker.fr/"
    }
  ]
}
```

BUG=#382

### DIFF
--- a/web-service/helpers.py
+++ b/web-service/helpers.py
@@ -172,6 +172,13 @@ def FetchAndStoreUrl(siteInfo, url, distance=None, force_update=False):
 ################################################################################
 
 def GetContentEncoding(content):
+    try:
+        # Don't assume server return proper charset and always try UTF-8 first.
+        u_value = unicode(content, 'utf-8')
+        return 'utf-8'
+    except UnicodeDecodeError:
+        pass
+
     encoding = None
     parser = lxml.etree.HTMLParser(encoding='iso-8859-1')
     htmltree = lxml.etree.fromstring(content, parser)
@@ -189,12 +196,8 @@ def GetContentEncoding(content):
             encoding = value[0]
 
     if encoding is None:
-        try:
-            encoding = 'utf-8'
-            u_value = unicode(content, 'utf-8')
-        except UnicodeDecodeError:
-            encoding = 'iso-8859-1'
-            u_value = unicode(content, 'iso-8859-1')
+        encoding = 'iso-8859-1'
+        u_value = unicode(content, 'iso-8859-1')
 
     return encoding
 


web-service/helpers.py
```diff
@@ -189,12 +196,8 @@ def GetContentEncoding(content):
             encoding = value[0]
 
     if encoding is None:
-        try:
-            encoding = 'utf-8'
-            u_value = unicode(content, 'utf-8')
-        except UnicodeDecodeError:
-            encoding = 'iso-8859-1'
-            u_value = unicode(content, 'iso-8859-1')
+        encoding = 'iso-8859-1'
```
dinhviethoa: It sounds like it might break for web sites encoded using utf-8 that don't advertise the charset.

dinhviethoa:
When I'm looking at the html of http://www.blackanddecker.fr, the charset is advertised.
I'm not sure why the code isn't catching it. We probably need to investigate that first.

```
<?xml version="1.0" encoding="iso-8859-1" ?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">

<html xmlns="http://www.w3.org/1999/xhtml" lang="fr">
<head><meta property="og:image" content="http://www.blackanddecker.co.uk/images/sharelogo.jpg" /><meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1" />
```

beaufortfrancois:
The problem here is the fact that http://www.blackanddecker.fr/ badly advertises its charset (as many other websites do). You can see that by running `curl www.blackanddecker.fr | xxd | less` and read:
```
00001b0: 2b44 4543 4b45 52e2 84a2 2065 7374 206c  +DECKER... est l
```

Here, `e2 84a2` is ™ UTF-8 Unicode. This means charset should be UTF-8 not ISO-8859-1.

By checking first if UTF-8 is the appropriate content encoding, we can deliver properly encoded `description` and `title`  values.

web-service/helpers.py
beaufortfrancois: Trying utf-8 encoding is done at the very beginning of `GetContentEncoding` function. That way, if a website is encoded using utf-8 but with a misconfigured charset, we can deliver right content.

web-service/helpers.py
```diff
@@ -189,12 +196,8 @@ def GetContentEncoding(content):
             encoding = value[0]
```
dinhviethoa: Therefore, I think we should try to decode with the encoding passed as parameter and invalidate it if it fails. The fallback on iso-8859-1 should always work.

dinhviethoa:
Ok. I'll add this improvement later then :)

web-service/helpers.py
mmocny: Sorry Hoa, I didn't see this comment before pulling.

We can revert, but I tested with the known test cases and this so far seems a strict improvement.  Can we do a subsequent patch if its still not perfect?

dinhviethoa: sounds good to me.

beaufortfrancois: For info, the encoding is not passed as parameter of  `GetContentEncoding`. 
What is passed is the page content which is used to detect the encoding of it.

PR #390, opened 2015-05-18
[PWS] Added Physical Web User Agent

beaufortfrancois:
Some websites rely on the "Mozilla5/0" part of the User-Agent string to serve web content. Let's add a Physical Web User Agent to fix that issue.

BUG=https://github.com/google/physical-web/pull/380


dinhviethoa:
- Do you have specific examples of website in mind?
- What do they serve if Mozilla5/0 is not available?

dinhviethoa:
It sounds like http://www.orange.fr is a good example.

PR #392, opened 2015-05-20
Let out favicon contentType check

beaufortfrancois:
Rather than restricting favicons strictly to `image/` contentType, let's add `application/octet-stream` so that it works on more websites.
Once again, http://www.orange.fr was failing...

### DIFF
--- a/web-service/helpers.py
+++ b/web-service/helpers.py
@@ -355,7 +355,8 @@ def StoreUrl(siteInfo, url, content, encoding):
             contentType = result.headers['Content-Type']
             if contentType is None:
                 icon = None
-            elif not contentType.startswith('image/'):
+            elif (contentType != 'application/octet-stream' and
+                  not contentType.startswith('image/')):
                 icon = None
     except:
         s_url = url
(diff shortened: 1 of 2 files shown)


mmocny:
Nice fix!

mmocny:
Hmm, `orange.fr` is still failing for me with this update..

mmocny:
..I mean, we are still not finding an icon, and so the tests fail with this update.

mmocny:
odd, now it works.  Perhaps temporary server outage?

beaufortfrancois:
For info, I also experienced network issues. Not sure if it comes from App Engine or where I'm from though.

beaufortfrancois:
Let me know when PWS is updated as well ;)

mmocny:
I've updated dev server, will do prod later today.

Quick question: do you think we should even be checking for the existence of the icon?  What do browsers do?  If the site has marked itself up to specify its icon, shouldn't we take it for its word?

beaufortfrancois:
I would personally remove the HEAD fetch for this specific reason.

mmocny:
#395

PR #393, opened 2015-05-20
[android] Use https for PWS

beaufortfrancois:
Based on #348 discussion, here's the patch to make android client use `https` instead of `http`.


cco3:
LGTM

PR #395, opened 2015-05-20
Remove icon HEAD fetch

mmocny:


mmocny:
This PR is based on a hash that will change, will need to rebase..

dinhviethoa:
The purpose was to prevent having URL of favicon that would be invalid / 404.

mmocny:
But the fallback is to return no url, which produces the same effect, right?

It does push to the client the act of fetching an invalid url, but I think this should happen infrequently in the wild.

And anyway, now that we route through the PWS for icon fetching, the bulk of the cost is on our server.

mmocny:
Specifically, clients will need to open 1 connection to PWS to make a resolve-url request, then make many requests over (hopefully) same connection for each icon.  If their is no icon to return, the added cost to the client is only the extra request payload on an already established tcp connection -- but now this happens async after the resolve-url, which is now unblocked on icon fetches.

I think thats a worthy tradeoff.

dinhviethoa:
Lgtm

PR #397, opened 2015-05-22
[android] Reflect new PWS metadata values

cco3:
This change properly uses the metadata values currently returned by PWS.

* Adds displayUrl
* Adds rank
* Removes score

@mmocny PTAL at PwsClient.java and make sure I'm not expecting any metadata values that aren't guaranteed, etc.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataComparator.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MetadataComparator.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2015 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


g-ortuno:
What's the difference between url and displayUrl?

cco3:
displayUrl is a new value returned by the metadata server that is the url that is supposed to be displayed.  I think the idea is that the url to be displayed could be a shortened url or a lengthened url depending on which one is more useful.

cco3:
lost a file...trying to find it...

g-ortuno:
But that's the same as siteUrl...

cco3:
Not necessarily given the description I just gave...displayUrl is a bit more variable.  See #388 for more info.

g-ortuno:
So we now have three urls?
1. id (basically the original URL) used for our map of devices
2. siteUrl used when opening a link
3. displayUrl self explanatory

g-ortuno:
LGTM

cco3:
That's correct.

PR #398, opened 2015-05-24
add other node.js examples

dermike:
Adding other examples heading to collect more examples of using node.js to scan for uribeacons, including an alternative example that also displays metadata using the PW proxy server.


scottjenson:
Excellent @dermike! Thank you

PR #404, opened 2015-05-27
[android] Fix sorting bug

cco3:
We had been using device addresses to do metadata lookups where urls
were the expected key.  As one might expect, this led to arbitrarily
sorted values.


mmocny:
I'm still seeing some odd sorting (sorts fine to a point, then there are more values after), but its probably unrelated to this patch.  Going to land and iterate.

Perhaps we show a sorted list, then append newly discovered uribeacons?

PR #405, opened 2015-05-27
[android] Fix unchecked ops and deprecations

cco3:
Additionally, add checks for these things to build.gradle.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
@@ -178,11 +178,13 @@ public class UriBeaconDiscoveryService extends Service
   }
 
   @Override
+  @SuppressWarnings("deprecation")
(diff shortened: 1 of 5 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
```diff
@@ -183,7 +183,7 @@ public int onStartCommand(Intent intent, int flags, int startId) {
     initializeLists();
     // Start scanning only if the screen is on
     PowerManager powerManager = (PowerManager) getSystemService(Context.POWER_SERVICE);
-    if (powerManager.isScreenOn()) {
+    if (powerManager.isInteractive()) {
```
g-ortuno: isInteractive is not available in API 19 which is our min SDK

cco3: good catch

cco3:
done

g-ortuno:
LGTM

PR #409, opened 2015-05-28
[android] Cleanup errs from debug output change

cco3:


mmocny:
Its time we update the PWS to link to the original url so that `url != displayUrl` and we can easier see the effects of these changes.

PR #412, opened 2015-05-29
add README's for iOS and Android clients with download links

dermike:
Adding README's with links to download respective iOS and Android clients. This makes them easier to find when navigating from the main README client links.


scottjenson:
Thanks!

PR #421, opened 2015-06-23
Debug data

dinhviethoa:
- Added debug information about delays,
- Removed URL expander when resolving the beacon and use only PWS data,
- Fixed Xcode7 warnings.

@mmocny

### DIFF
--- a/ios/PhyWeb/Backend/PWBeaconManager.m
+++ b/ios/PhyWeb/Backend/PWBeaconManager.m
@@ -327,9 +327,6 @@
   if (idx != NSNotFound) {
     NSDictionary* beaconInfo = [_pendingBeaconsInfos objectAtIndex:idx];
     NSTimeInterval discoveryDelay = [beaconInfo[DISCOVERY_DELAY_KEY] doubleValue];
-    // NSTimeInterval requestStartTime = [beaconInfo[@"requestStartTime"]
-    // doubleValue];
-    // NSTimeInterval currentTime = [NSDate timeIntervalSinceReferenceDate];
     NSTimeInterval requestDelay = [request delay];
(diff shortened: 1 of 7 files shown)


ios/PhyWeb/Backend/PWBeaconManager.m
```diff
@@ -113,19 +117,20 @@ - (void)unregisterConfigurationChangeBlock:(id)registeredBlock {
...
+    PWBeaconManager* strongSelf = weakSelf;
+    [strongSelf _updateBeacons];
   }];
   _httpServiceBrowser = [[NSNetServiceBrowser alloc] init];
   [_httpServiceBrowser setDelegate:self];
   _httpsServiceBrowser = [[NSNetServiceBrowser alloc] init];
   [_httpsServiceBrowser setDelegate:self];
-  [_httpServiceBrowser searchForServicesOfType:@"_http._tcp." inDomain:nil];
-  [_httpsServiceBrowser searchForServicesOfType:@"_https._tcp." inDomain:nil];
+  [_httpServiceBrowser searchForServicesOfType:@"_http._tcp." inDomain:@""];
```
mmocny: What is the implication of this change?  Was this the "warnings" fix?

mmocny:
Looks good.  I can't quite grok what the code block dealing with !hasDisplayUrl using PWURLShortener code was doing -- but seeing as it's being removed, thats fine :P

ios/PhyWeb/Backend/PWBeaconManager.m
dinhviethoa: Yes. That was a warning fix. The second parameter of -searchForServicesOfType:inDomain: can't be nil. The correct value is an empty string.

PR #422, opened 2015-06-23
[android] Use setServiceUuid to build scan filter

cco3:
Not only is this simpler, but this ensures that other mask values, etc.
are set to null and not an empty byte array.  This ensures that all the
appropriate beacons show up and fixes a bug that hid some in the
surfaced notifications.


cco3:
This fixes the bug that kept the parking demo from showing up in the notifications.

g-ortuno:
Nice! How come this didn't work before?

cco3:
I assume that there's something specific the filter code looks for when you use an empty byte array instead of null.  I don't really understand the details of what a "servicedatamask" is to know what kinds of values can be masked.

PR #426, opened 2015-06-25
[android] Create a PwoMetadata class

cco3:
PWO = Physical Web Object
This is a generic concept that includes BLE devices, MDNS, SSDP, etc.
Having a class to manage all data related to the physical web object
itself and not just its url metadata will simply lookups of this
information.


matthewsibigtroth:
LGTM. Waiting for Travis...

PR #429, opened 2015-06-25
[android] Use PwoMetadata in the ListView adapter

cco3:
Instead of keeping a list of device addresses, we hang onto a list of
PwoMetadata objects to simplify things a little.


matthewsibigtroth:
LGTM. Looking at you Travis...

PR #430, opened 2015-06-25
[android] Use PwoMetadata in DiscoveryService

cco3:
Simplify the code some by using the PwoMetadata class in the
UriBeaconDiscoveryService.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
```diff
@@ -356,14 +364,20 @@ private void updateNotifications() {
   }
 
   private ArrayList<String> getSortedBeaconsWithMetadata() {
-    ArrayList<String> unSorted = new ArrayList<>(mDeviceAddressToUrl.size());
+    ArrayList<PwoMetadata> pwoMetadataList = new ArrayList<>(mUrlToPwoMetadata.size());
```
matthewsibigtroth: do we need the size() call here?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
```diff
@@ -356,14 +364,20 @@ private void updateNotifications() {
...
+      PwoMetadata pwoMetadata = mUrlToPwoMetadata.get(mDeviceAddressToUrl.get(key));
+      if (pwoMetadata != null && pwoMetadata.hasUrlMetadata() && pwoMetadata.hasBleMetadata()) {
+        pwoMetadataList.add(pwoMetadata);
       }
     }
-    Collections.sort(unSorted, new MetadataComparator(mUrlToUrlMetadata));
-    return unSorted;
+    Collections.sort(pwoMetadataList);
+
+    ArrayList<String> deviceAddresses = new ArrayList<>(mUrlToPwoMetadata.size());
```
matthewsibigtroth: do we need the size() call here?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeaconDiscoveryService.java
cco3: No, it's just an algorithmic optimization.  same for the below.

matthewsibigtroth:
Two minor comments. Otherwise LGTM.

cco3:
If that's it, I'm going to leave the size parameter and submit.

PR #432, opened 2015-06-26
[android] Rename UriBeacon Service to Pwo Service

cco3:
UriBeaconDiscoveryService -> PwoDiscoveryService
There's several reasons to rename this class, but we are already
finding non-uribeacon PWOs (i.e. via MDNS, SSDP).


matthewsibigtroth:
LGTM

PR #433, opened 2015-06-26
[android] Use PwoMetadata as request object

cco3:
Instead of passing multiple attributes to PwsClient, this change gives
it a single PwoMetadata object.  This will especially simplify things
if we eventually decide to send more data up to PWS.


matthewsibigtroth:
LGTM

PR #434, opened 2015-06-26
[android] Stop storing device addresses in Service

cco3:
The PwoDiscoveryService is needlessly storing device addresses.  This
change corrects that behavior.


matthewsibigtroth:
LGTM

PR #435, opened 2015-06-29
Show urls in batches

cco3:
The end goal of this set of commits is to mimic our behavior on iOS of showing batches of urls at a time.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
@@ -85,7 +85,6 @@ public class NearbyBeaconsFragment extends ListFragment
   private HashMap<String, PwoMetadata> mUrlToPwoMetadata;
   private long mScanStartTime;
   private AnimationDrawable mScanningAnimationDrawable;
-  private boolean mIsScanRunning;
   private Handler mHandler;
   private NearbyBeaconsAdapter mNearbyDeviceAdapter;
   private Parcelable[] mScanFilterUuids;
@@ -97,10 +96,7 @@ public class NearbyBeaconsFragment extends ListFragment
   private Runnable mScanTimeout = new Runnable() {
     @Override
     public void run() {
-      mScanningAnimationDrawable.stop();
-      scanLeDevice(false);
-      mMdnsUrlDiscoverer.stopScanning();
-      mSsdpUrlDiscoverer.stopScanning();
+      stopScanning();
       mNearbyDeviceAdapter.sortUrls();
       mNearbyDeviceAdapter.notifyDataSetChanged();
       fadeInListView();
@@ -152,7 +148,6 @@ public class NearbyBeaconsFragment extends ListFragment
     TextView tv = (TextView) rootView.findViewById(android.R.id.empty);
     //Get the top drawable
     mScanningAnimationDrawable = (AnimationDrawable) tv.getCompoundDrawables()[1];
-    mScanningAnimationDrawable.start();
   }
 
   public View onCreateView(LayoutInflater layoutInflater, ViewGroup container, Bundle savedInstanceState) {
@@ -166,20 +161,13 @@ public class NearbyBeaconsFragment extends ListFragment
     super.onResume();
     getActivity().getActionBar().setTitle(R.string.title_nearby_beacons);
     getActivity().getActionBar().setDisplayHomeAsUpEnabled(false);
-    mScanningAnimationDrawable.start();
-    scanLeDevice(true);
-    mMdnsUrlDiscoverer.startScanning();
-    mSsdpUrlDiscoverer.startScanning();
+    startScanning();
   }
 
   @Override
   public void onPause() {
     super.onPause();
-    if (mIsScanRunning) {
-      scanLeDevice(false);
-      mMdnsUrlDiscoverer.stopScanning();
-      mSsdpUrlDiscoverer.stopScanning();
-    }
+    stopScanning();
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -213,28 +235,38 @@ public void onUrlMetadataIconReceived(PwoMetadata pwoMetadata) {
...
+    mMdnsUrlDiscoverer.stopScanning();
+    mSsdpUrlDiscoverer.stopScanning();
+    mBluetoothAdapter.stopLeScan(mLeScanCallback);
+    // Change the display appropriately
+    mSwipeRefreshWidget.setRefreshing(false);
+    mScanningAnimationDrawable.stop();
+  }
+
+  @SuppressWarnings("deprecation")
+  private void startScanning() {
```
matthewsibigtroth: Perhaps we could add a short description of the scanning process, particularly with respect to the three timeouts?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -93,19 +95,46 @@
...
     public void run() {
-      mScanningAnimationDrawable.stop();
-      scanLeDevice(false);
-      mMdnsUrlDiscoverer.stopScanning();
-      mSsdpUrlDiscoverer.stopScanning();
-      mNearbyDeviceAdapter.sortUrls();
-      mNearbyDeviceAdapter.notifyDataSetChanged();
-      fadeInListView();
+      Log.d(TAG, "running first scan timeout");
+      if (!mPwoMetadataQueue.isEmpty()) {
```
matthewsibigtroth: If this line is true, does this mean that the ui will indicate scanning is complete, while scanning is still running?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -254,14 +286,8 @@ private boolean leScanMatches(ScanRecord scanRecord) {
 
   @Override
   public void onRefresh() {
```
matthewsibigtroth: Do we need to remove any potential callbacks here before starting another scan?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
cco3: It means it will start showing results while more may come in, like it already currently.

cco3:
k, done.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
cco3: done.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
matthewsibigtroth: Ok. I guess my only concern was that the user might think scanning is done, while scanning is actually occurring. The only solution I can think of is that we keep the refresh animation going until the third timeout, but then it's blocking the top of the list view, which seems suboptimal. I guess as long as we have parity with iOS we should be ok  :)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
matthewsibigtroth: Nice!

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -192,7 +221,7 @@ public void onPrepareOptionsMenu(Menu menu) {
   @Override
   public void onListItemClick(ListView l, View v, int position, long id) {
     // If we are scanning
-    if (mIsScanRunning) {
+    if (mScanningAnimationDrawable.isRunning()) {
```
matthewsibigtroth: It's possible that the animation may not be animating, but scanning is still happening. This might not matter after the first timeout, since new items are appended to the list, so perhaps this is moot?

cco3: Right, the animation should only be running when there are no items in the list.  The full scan goes for 10 seconds now, not just 3, so we need to let people interact with the listview before then.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -213,28 +242,38 @@ public void onUrlMetadataIconReceived(PwoMetadata pwoMetadata) {
...
+    mBluetoothAdapter.stopLeScan(mLeScanCallback);
+    // Change the display appropriately
+    mSwipeRefreshWidget.setRefreshing(false);
+    mScanningAnimationDrawable.stop();
+  }
+
+  @SuppressWarnings("deprecation")
+  private void startScanning() {
+    // Clear any stored url data
+    mUrlToPwoMetadata.clear();
```
matthewsibigtroth: Do we also want mPwoMetadataQueue.clear() here?

cco3: done

matthewsibigtroth:
LGTM

PR #436, opened 2015-06-30
Conform scanning across mediums

cco3:
This series of commits creates a new PwoDiscoverer class and conforms the different discoverers to this one interface.


matthewsibigtroth:
Looking slick! And the plot thickens as we approach the vaunted uniform scanner. LGTM.

PR #437, opened 2015-06-30
[android] Further fortify mdns start/stop

cco3:
The mdns scan start/stop still isn't robust enough.  The reason seems to be
that while we are waiting on a response, client code can call start/stop a
second time and the NsdManager doesn't like that.


matthewsibigtroth:
LGTM

PR #441, opened 2015-07-01
Remove UrlShortener class

cco3:
The end goal of these commits is to remove the UrlShortener class and rely on the PwsClient instead.


matthewsibigtroth:
LGTM

PR #442, opened 2015-07-01
[android] Use ble discoverer in discovery service

cco3:
There's significant behavior change here in that instead of a low power
scan that runs constantly while the screen is on, we instead do a high
power scan that runs for 10 seconds total on screen wakeup.


cco3:
K, this is on hold until #443 goes in.

cco3:
rebased on top of #443

matthewsibigtroth:
LGTM

PR #443, opened 2015-07-01
[android] Handle screen on in autostart receiver

cco3:
Instead of creating a receiver and filtering screen events at runtime,
we bind to these at the manifest level.


cco3:
yea, I also don't think there's a problem with start and stop being called at the same time.

matthewsibigtroth:
LGTM

Issue #445, opened 2015-07-02
Awkward treatment of urn:uuid beacons in the Physical Web scanner (Android)


devunwired:
The current implementation of the PWO scanning features leaves a strange loophole in detecting URN:UUID broadcasters. This is the result of two event sequences:

1. The `NearbyBeaconsFragment.onPwoDiscovered()` [implementation](https://github.com/google/physical-web/blob/73d00f2e40a2449651148d8d5d1ae7c4be57f6eb/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java#L266) (after calling through to `emptyPwoMetadataQueue()`) adds all valid beacons to the `NearbyBeaconsAdapter`, regardless of URI type. However, `notifyDataSetChanged()` is not triggered at this point, so the detected items are not displyed in the UI.

2. At a later time, the `onUrlMetadataReceived()` [implementation](https://github.com/google/physical-web/blob/73d00f2e40a2449651148d8d5d1ae7c4be57f6eb/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java#L212) is called when the `PwsClient` returns description information about the beacon's URL. When this event occurs, `notifyDataSetChanged()` is triggered and the list updates.

In the above sequence, beacons with a valid `http(s)` URL will eventually pass both and display. However, `urn:uuid` beacons will only pass the first test as the service will return no metadata for them (an expected behavior).  This leads to two conflicting conditions:

* When only `urn:uuid` beacons are present, the list will never show them.
* When a mix of `urn:uuid` and `http(s)` beacons are present, the list will show all of them. In this case, the `urn:uuid` beacons launch a blank web page.

I'm not certain whether the goal of the physical web application is to hide or show the beacons, a proposed fix would depend on the intended behavior here.

devunwired:
Come to think of it, this would keep `http(s)` beacons from showing up in the list as well if the `PwsClient` were unable to reach the Physical Web Service and return valid metadata for at least one of them…

cco3:
>  However, notifyDataSetChanged() is not triggered at this point

It should be.  This is a *very* recent refactoring if you hadn't noticed.  Thanks for the great catch.

Assuming 1&2 need to happen for the issue to happen, I've only glanced at the rest of your issue.  Please let me know if the PR I just made isn't enough.

devunwired:
[...]

PR #446, opened 2015-07-06
[android] Update UI when emptying metadata queue

cco3:
The whole intend of queueing up metadata before emptying into in the
adapter is so that we can control when the information is displayed.

This fixes issue #445


matthewsibigtroth:
LGTM

PR #447, opened 2015-07-06
Remove IDE project files from repo.

devunwired:
IML files for Android Studio are generated files and should not be checked into source. They are always derived from the `build.gradle` files which represent the source of truth in the project. Their format is subject to variation as different versions of Studio will generate a different XML structure in the files.

This commit removes the checked-in files and adds the type to `.gitignore`.


g-ortuno:
See #371

g-ortuno:
Do you have a source saying that this files are not useful?

devunwired:
@g-ortuno Certainly, see below…

Comments from an Android Studio tools team developer indicating they are generated content:
https://code.google.com/p/android/issues/detail?id=77983#c7
https://code.google.com/p/android/issues/detail?id=156708#c6

StackOverflow answer form another tools team developer indicating this + the entire `.idea/` directory should not be tracked in a Studio project:
http://stackoverflow.com/a/27570950/246461

…If I can find the Google I/O video where they made the same statement on stage, I'll post it here as well.

cco3:
LGTM.

g-ortuno:
@cco3 beat me to the lgtm

PR #450, opened 2015-07-06
[android] Enforce http(s) only on BLE beacons discovered.

devunwired:
This commit uses the same `URLUtil.isNetworkUrl()` logic as the rest of the application to ensure that only `http` and `https` prefixed UriBeacons are listed in the nearby list, per the discussion in issue https://github.com/google/physical-web/issues/445.

This check replaces the null/empty check, because `isNetworkUtil()` does that for us also.


cco3:
Thanks!

PR #453, opened 2015-07-06
Added charts for distance and RSSI

dinhviethoa:
Added charts for distance and RSSI for Physical Web beacons.
It will be available when debug mode is turned on.
It's triggered when choosing any beacon in the list.

review: @mmocny


dinhviethoa:
TBR.

PR #455, opened 2015-07-07
[android] Refresh without killing DiscoveryService

cco3:
This change will also prepare us for when the service chaches
PwoMetadata objects because at that point merely killing the
service won't be enough to force a refresh.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
@@ -139,36 +139,39 @@ public class NearbyBeaconsFragment extends ListFragment
    */
   private class DiscoveryServiceConnection implements ServiceConnection {
     private PwoDiscoveryService mDiscoveryService;
+    private boolean mRequestCachedPwos;
 
     @Override
-    public void onServiceConnected(ComponentName className, IBinder service) {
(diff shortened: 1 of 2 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -139,36 +139,39 @@ public boolean onItemLongClick(AdapterView<?> av, View v, int position, long id)
    */
   private class DiscoveryServiceConnection implements ServiceConnection {
     private PwoDiscoveryService mDiscoveryService;
+    private boolean mRequestOldPwos;
```
matthewsibigtroth: I know this is not completely defined yet, but what are we considering to be old pwos? Those discovered on the most recent scan?

matthewsibigtroth: Ah, it's whatever is currently in mUrlToPwoMetadata, right?

cco3: Yea, maybe I should call this mRequestCachedPwos

cco3: ok, updated the variable name

matthewsibigtroth:
LGTM

PR #456, opened 2015-07-07
[android] Require explicit call to DL favicon

cco3:
This is desirable for three reasons:

1. These are independent features of the PWS, the calls should be made
   distinct
2. The BeaconConfigFragment doesn't use the icon, so we shouldn't
   download it
3. When we start caching PWOs in the discovery service, we'll need to
   fetch icons if we already have the url metadata, but no icon.


matthewsibigtroth:
Could we not include the favicon as part of the metadata returned from PWS (e.g. base64 encoded)? Then the PWS would only fetch the favicon string if the requester asked for it?

cco3:
A variant of this was discussed in #362.  We'd like to be able to show the entry before necessarily having the favicon.  Maybe we can get the PWS fast enough that it wouldn't matter though, dunno.

matthewsibigtroth:
LGTM

PR #459, opened 2015-07-08
[android] Returned stored PwoMetadata to clients

cco3:
Currently, when the Fragment refreshes, we hand it newly discovered
PwoMetadata objects.  However, because the service (typically) already
has the UrlMetadata and the icon, this can prevent the fragment from
being subscribed to the UrlMetadata on refresh.  Instead of giving the
client the newly discovered PwoMetadata objects, we give it the stored
one that matches the url of the newly discovered one.


matthewsibigtroth:
LGTM

PR #462, opened 2015-07-08
[android] Handle scanning startup in onCreate

cco3:
onStartService isn't the place to do this since we only want to start
scanning once.  The android docs seem to indicate that onCreate doesn't
have to be merely used for variable instantiation, but can also start
threads as shown in one of their own examples.

http://developer.android.com/guide/components/services.html


matthewsibigtroth:
LGTM

PR #463, opened 2015-07-08
[android] Persist PwoMetadata after service death

cco3:
This change keeps around some PwoMetadata so that if a user opens the
app sometime after the service dies, it can still start quickly.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoDiscoveryService.java
@@ -21,7 +21,9 @@ import org.physical_web.physicalweb.PwoMetadata.UrlMetadata;
 import android.app.Notification;
 import android.app.PendingIntent;
 import android.app.Service;
+import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Color;
@@ -35,11 +37,15 @@ import android.util.Log;
 import android.view.View;
 import android.widget.RemoteViews;
 
+import org.json.JSONException;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoDiscoveryService.java
```diff
@@ -63,6 +69,11 @@
 
   private static final String TAG = "PwoDiscoveryService";
   private static final String NOTIFICATION_GROUP_KEY = "URI_BEACON_NOTIFICATIONS";
+  private static final String SHARED_PREFS_FILE = "pwo_discovery_service_cache";
```
matthewsibigtroth: Would it be useful to consolidate this preference file with that used in MainActivity?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoDiscoveryService.java
```diff
@@ -227,7 +291,17 @@ public void onPwoDiscovered(PwoMetadata pwoMetadata) {
     PwoMetadata storedPwoMetadata = mUrlToPwoMetadata.get(pwoMetadata.url);
     if (storedPwoMetadata == null) {
```
matthewsibigtroth: If storedPwoMetadata is null, then pwoResponseCallback.onPwoDiscovered(storedPwoMetadata) is still called later in this method, receving null as the argument.

matthewsibigtroth: Wait, scratch that, didn't see the storedPwoMetadata = pwoMetadata line. :)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoDiscoveryService.java
```diff
@@ -141,6 +153,38 @@ private void initialize() {
     mHandler = new Handler();
     mUrlToPwoMetadata = new HashMap<>();
     mCanUpdateNotifications = false;
+
+    // Make sure we are trying to load the right version of the cache
```
matthewsibigtroth: initialize and onDestroy are starting to get a little beefy, so we might want to break them out into other methods at some point.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoDiscoveryService.java
cco3: Dunno...lemme check best practices...

I think it makes sense to keep it distinct, though we should be using better scoped names.

Let's do #465 first, and then I'll change this.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoDiscoveryService.java
cco3: done

matthewsibigtroth:
LGTM

PR #465, opened 2015-07-10
[android] Use a scoped preference file name

cco3:
"""
When naming your shared preference files, you should use a name that's
uniquely identifiable to your app, such as
"com.example.myapp.PREFERENCE_FILE_KEY"
"""
http://developer.android.com/training/basics/data-storage/shared-preferences.html


matthewsibigtroth:
LGTM

PR #466, opened 2015-07-13
match new API that supports extensions

dinhviethoa:
Since we can't refer to `[UIApplication sharedApplication]` in extensions, the API implementation of UriBeacon has been modified (https://github.com/google/uribeacon/pull/256) to match that and then, Physical Web needs the following changes.

@mmocny


mmocny:
lgtm

PR #468, opened 2015-07-13
[android] Get rid of flashing scanning animation

cco3:
If we load cached results, we don't need to show the scanning
animation.  This change makes sure that animation doesn't get
shown if it's not needed.


cco3:
Updated to fix bug on first use.

matthewsibigtroth:
LGTM

PR #469, opened 2015-07-14
[android] Remove uribeacon library services

cco3:
The uribeacon library services are no longer needed.


g-ortuno:
Don't you need these for configuration?

cco3:
Oh duh...I was wondering what the gatt service was for.

PR #471, opened 2015-07-14
[android] Create a service to listen to the screen

cco3:
We have switched from having the PwoDiscoveryService be a long running
service, but SCREEN_ON and SCREEN_OFF don't actually trigger when set
in the manifest.  Therefore, let's create a second service that will
monitor these things.  Since the code is isolated, it will be easy to
remove if/when we come up with a better solution


matthewsibigtroth:
LGTM

PR #473, opened 2015-07-14
[android] Use start/end instead of begin/end

cco3:
start/end is considered better because it enables bidi support.


matthewsibigtroth:
LGTM

Issue #452, opened 2015-07-06
Physical Web Service Link Follows


devunwired:
This is more of a higher-level "what-if" discussion regarding the uses of the Physical Web Service:
https://github.com/google/physical-web/tree/master/web-service

Using the PWS as a trusted intermediary through which content requests from passers by can be made for gathering site metadata is a great idea, and it makes perfect sense. At that point, the user hasn't necessarily engaged the item so we don't want to provide any way of tracking them.

However, in the current implementation of the mobile scanners, the PWS also acts as a "go" proxy for following the beacon's link. I'm not so certain this is as helpful. Once a user chooses to engage with the beacon, the proxy is not providing as much help to them.

On the positive side, the PWS could be used to blacklist URLs and protect users from following known malicious links (even if they try). However, the problem that is introduced (at least on mobile) is that the proxy interrupts the deep linking experience because it forces the browser (typically Chrome) to launch in order to redirect to the actual advertised URI.

If that advertised URI is intended to be captured as a deep link into a native application, the intermediate launching of the browser is unnecessary. This experience is visually jarring on both platforms due to the multiple app switching animations. It is a slightly more so on Android due to the potential for Intent resolution dialogs at each turn asking the user which application should handle the link.

I realize the PWS is not a necessary part of the overall project, but as an example I wonder if it shouldn't showcase how to get the best experience across all platforms. On mobile, that means best practices around deep links.

cco3:
We were actually just discussing what we wanted to do about the redirecter.  The native application link is a really good point.  It might be a while before we take any action on this, but it's something on our minds at the moment and this gives some extra perspective.

mmocny:
Great question.  We are actually also deliberating the pros and cons of this usage.

I will point out that most Physical Web urls today point to a commodity url shortener, and so the experience will still be subpar even if we remove the go link.  Since most shorteners request that they are always used (must-revalidate), I think it not a solution to just link to the final url.

However, if you roll your own shortener on your own domain, and you handle deep links within your app, this would work fine.

Great additional data point!

devunwired:
One thing I'm curious to dig into more is how the updates to the goo.gl shortener with app indexing work:
http://googlewebmastercentral.blogspot.com/2015/05/app-deep-linking-with-googl.html

[...]

PR #476, opened 2015-07-15
[android] Drop go link redirection

cco3:


dinhviethoa:
lgtm.

mmocny:
Late to the party, but, huzzah!

Also: #452 is relevant.  We won't be able to close that one quite yet, since there are a few other tasks and we will need to wait for all clients to get updated (may take a while).

PR #478, opened 2015-07-20
[android] remove some redundant code

krpiotrek:


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -300,10 +300,8 @@ private void handleFoundDevice(final ScanResult scanResult, ParcelUuid filteredU
       scanLeDevice(false);
       mUriBeaconConfig = new UriBeaconConfig(getActivity(), new UriBeaconConfigCallback(),
                                              filteredUuid);
-      if (mUriBeaconConfig != null) {
```
g-ortuno: I don't think you can do this. There are many cases in which mUriBeaconConfig could be null.

krpiotrek: It's being created one line above, and it's never nulled - am I missing something?

g-ortuno: Ah you're right. Sorry, got confused with `ConfigUriBeacon.createConfigUriBeacon()`

cco3:
Thanks!

PR #484, opened 2015-08-03
[android] Cancel prev. notifications on new scan

cco3:
This restores previous behavior and fixes a bug.  Currently, if
a user encounters 2+ URLs and then later encounters exactly 1 URL,
two notifications are displayed.  If we kill all the notifications
at the beginning of a scan, we remove the risk of this happening.


cco3:
After this change, we can make a new release

matthewsibigtroth:
LGTM

matthewsibigtroth:
LGTM sans Close

PR #487, opened 2015-08-04
[android] Restart scanners on refresh

cco3:
This fixes a bug with refreshing mdns results, since mdns results
are typically only found once in a scan, and not frequently like ble
results.


cco3:
@matthewsibigtroth, don't accept this just yet, but my suspicion is that this is the issue you were running into.

cco3:
Alright @matthewsibigtroth, this is ready to go.

matthewsibigtroth:
LGTM

PR #489, opened 2015-08-07
Grammar/wording change

mozz100:
Nothing major, just the words jarred when I first read them. I think this is more appealing.


scottjenson:
thanks!

PR #492, opened 2015-08-13
[android] Get urls from MDNS TXT record

cco3:
This change deprecates broadcasting a url as the MDNS service name.
Instead, if the service type is "physical-web", it will grab a JSON
array of urls from the "urls" attribute of that service.

### DIFF
--- a/documentation/mDNS_Support.md
+++ b/documentation/mDNS_Support.md
@@ -1,5 +1,7 @@
 ## mDNS Support
 
+**Due to several issues, mDNS support is currently disabled this app!  We intend to revisit (and hopefully reenable) this feature in the future.**
+
 The Physical Web is about getting URLs into the physical world. However, this isn't limited to just Bluetooth Low Energy (BLE) beacons. [mDNS](http://en.wikipedia.org/wiki/Multicast_DNS) is a service broadcast technique used in Wi-Fi networks. It has two advantages over BLE: 
 
 1. Only people logged into your Wi-Fi can see the mDNS URLs. This means that in an apartment, your neighbors can't see your devices.


maggu2810:
Should I reply here or would you want to keep the thread at the JmDNS issue?

This is also "just" an implementation of the discovery for android devices.
IMHO it would be better to start with an explanation what you try to do using DNS-SD and start with a specification at all. The implemetation should be the next step, after the specification is finished.

Are the hosts different from publisher and target? Or why you can't use http service?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsPwoDiscoverer.java
```diff
@@ -33,15 +40,64 @@ public void onDiscoveryStarted(String regType) {
...
+      }
+
+      // Parse out the array of urls
+      Map<String, byte[]> attributeMap = service.getAttributes();
+      byte[] urlsBytes = attributeMap.get("urls");
+      if (urlsBytes == null) {
+        Log.i(TAG, "`urls` attribute misisng");
+        return;
+      }
+      String urlsString = new String("urlsBytes");
```
matthewsibigtroth: Should this instead be:

String urlsString = new String(urlsBytes);

cco3:
This is just a prototype app where we do quick experiments, get things wrong, and hopefully the community to help us get it right.  So that's why you don'

cco3:
@maggu2810 This is just a prototype app where we do quick experiments, get things wrong, and hopefully the community helps us get it right. So that's why there's not a design spec for everything.

Anyway, our goal here is to use MDNS to advertise urls.  What do the publisher and target refer to in this context?  What do you mean by using an http service?  What do you think the right way of broadcasting urls over MDNS would be?

maggu2810:
@cco3 

I already posted some related documents in the referenced topic above (openhab/jmdns#25). Please read the mDNS stuff.
I can't read them for you, no remaining time in my spare time. Sorry, but I am very busy at the moment.

Here some document references for the specific topics.
* http://www.multicastdns.org/
* http://www.dns-sd.org/

> What do the publisher and target refer to in this context?

* I said publisher to the one, that publishs an serivce through multicast.
* target is a field in the DNS SRV Ressource Record (https://www.ietf.org/rfc/rfc2782.txt page 3)

What do you mean by using an http service?

* There is already a service type for http URLs. (http://www.dns-sd.org/servicetypes.html)
<pre>
http            World Wide Web HTML-over-HTTP
                Tim Berners-Lee <timbl at W3.org>
                Protocol description: RFC 2616
                Defined TXT keys: u=<username> p=<password> path=<path to document>
                (see txtrecords.html#http)
                Known Subtypes: _printer
                NOTE: ...
</pre>
This is just the first part. See the homepage for more information. There is also other another documents linked, for the TXT record spec of the http service, etc.
http://www.dns-sd.org/txtrecords.html#http

> What do you think the right way of broadcasting urls over MDNS would be?

This will depend on what aspects I have to fulfill.
I have to read the specs, too and have a look at
* if there is still something present, that could be used for my use case
* if there is nothing, that could be used, I would create a new service type and have to read all the specs to do it right, compliant, ...

scottjenson:
> What do you think the right way of broadcasting urls over MDNS would be?
>> This will depend on what aspects I have to fulfill.
>> I have to read the specs, too and have a look at

This is the question that concerns us. If you have a concrete proposal on how to improve what we're proposed, we're all ears.

cco3:
Hmm...the urlbookmark service type does look quite fitting.  I've uploaded a new patchset that uses that.

documentation/mDNS_Support.md
```diff
@@ -20,17 +20,18 @@ Our sample file `physical-web-url.service` looks like this:
 <!DOCTYPE service-group SYSTEM "avahi-service.dtd">
 
 <service-group>
-  <name replace-wildcards="yes">http://www.mycompany.com/xyz.html</name>
+  <name replace-wildcards="yes">Physical Web on %h</name>
   <service>
-    <host-name>www.mycompany.com</host-name>
-    <type>_http._tcp</type>
+    <type>_urlbookmark._tcp</type>
+    <host-name>local</host-name>
```
maggu2810: Don't think you need to set host-name at all.
Using this line all your FQDN are equal to local, this seems not correct.

<pre>man avahi.service</pre>
&lt;host-name&gt;  The  host name of the host that provides this service. This should be a host that is resolvable by multicast or unicast DNS. Please note that you need to specify a fully-qualified domain name (FQDN) here, i.e. .local is not appended implicitly! The  host  name  doesn't need to be part of the domain specified in &lt;domain-name&gt;. See avahi.hosts(5) for more information how to publish additional host name mappings.

documentation/mDNS_Support.md
```diff
@@ -20,17 +20,18 @@ Our sample file `physical-web-url.service` looks like this:
...
     <port>80</port>
-    <txt-record>path=/xyz.html</txt-record>
+    <txt-record>URL=http://www.mycompany.com/xyz.html</txt-record>
+    <txt-record>name=XYZ for great profit</txt-record>
   </service>
 </service-group>
 ```
 
-Then `<name>` tag must be unique on your network. We are suggesting that you use the URL as your name to be safe. 
+Then `<name>` tag must be unique on your network.
```
maggu2810: Why?

documentation/mDNS_Support.md
```diff
@@ -20,17 +20,18 @@ Our sample file `physical-web-url.service` looks like this:
 <!DOCTYPE service-group SYSTEM "avahi-service.dtd">
 
 <service-group>
-  <name replace-wildcards="yes">http://www.mycompany.com/xyz.html</name>
+  <name replace-wildcards="yes">Physical Web on %h</name>
   <service>
-    <host-name>www.mycompany.com</host-name>
-    <type>_http._tcp</type>
+    <type>_urlbookmark._tcp</type>
```
dinhviethoa: hmmm what was the issue with _http._tcp?

maggu2810: Don't know, because I don't know what you want to do with this project.
But if the publisher of the service (or let's they the target field of the DNS SD RR) does not fit the host, you cannot use it.
This has been pointed by the discussion before: http://www.dns-sd.org/txtrecords.html#http

cco3: Hoa, We were shoving the url in the service name, which that field is not designed for.  _urlbookmark is a registered service type which includes txt records with the keys `URL` and `name` like we would want.

dinhviethoa: The URL was split into the field 'type' (http vs https), host-name (www.mycompany.com), port (80) and path (/xyz.html), which result in http://www.mycompany.com:80/xyz.html
The name was just to help people choose a unique name and could be anything else.

cco3: Oh, hm, I see...yea, I guess we should just have a different suggested name then and possibly leave it as is.  Although...the urlbookmark service type seems to be directly targeted at what we are doing.  I'll have to look around a bit more.

PR #503, opened 2015-09-06
add physical-web-scan link as discussed in c5d3e1d

dermike:
@scottjenson suggestion for physical-web-scan description + some fixes of the node-eddystone-beacon-scanner description.


scottjenson:
Ah, thank you @dermike I was clearly a bit confused. Appreciate you fixing that.

PR #507, opened 2015-09-14
[android] add group to UrlMetadata and debug view

nondebug:
This adds the "group" key to UrlMetadata and displays it in debug view.  Currently the PWS doesn't return any groups so it should be empty.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoMetadata.java
```diff
@@ -139,6 +139,7 @@ public String getRegionString() {
     private static final String ICON_URL_KEY = "iconUrl";
     private static final String ICON_KEY = "icon";
     private static final String RANK_KEY = "rank";
+    private static final String GROUP_KEY = "group";
```
mmocny: Based on #508 the key is `"groupid"`

PR #511, opened 2015-09-16
s/group/groupid/

nondebug:


mmocny:
lgtm.

The PWS change has landed but hasn't been deployed.  I'll push a version to the dev server, so we can play with this.

PR #513, opened 2015-09-18
Expose groupid in Android client

nondebug:
* Fix the groupid key for incoming PWS messages ('group' -> 'groupid')
* Truncate the groupid to 4 characters for debug display

Note: groupid is currently only supported with the DEV endpoint url.
Caching behavior will cause the field to be blank in most cases as the
client defaults to PROD until debug mode is enabled.


nondebug:
Ellipsize is cleaner and allows us to show more of the groupid when we can, thanks for the suggestion.

cco3:
LGTM

nondebug:
Also renamed variables containing "group" to contain "groupid" instead

PR #516, opened 2015-09-21
Deduplicate grouped PWOs (PWOs with matching domain and page title)

nondebug:


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -389,7 +389,47 @@ private void safeNotifyChange() {
     }
 
     public void addItem(PwoMetadata pwoMetadata) {
-      mPwoMetadataList.add(pwoMetadata);
+      // If multiple PWOs have the same domain and page title, we consider them to be in the same
```
mmocny: Nit: Replace "same domain and page title" with "same groupid"

mmocny: ..and instead of describing the specific set of actions, consider describing the reason we are doing this:

"Multiple urls from the same group basically represent the same result, but with minimally different URLs (e.g. like a differing hash variable).  We would like to not display them all and confuse the user -- instead, we display one grouped result, and take the user to URL stored on the closest beacon when they click"

mmocny:
I worry that this implementation will be brittle.  Perhaps we need to more formally introduce the concept of a group.

Perhaps the fragment should not accept `PwoMetadata` objects any more, but instead accept a new `PwoGroup` (or something), which is a collection of `PwoMetadata`, and display the first (closest) result.  On click, we look up the group that was clicked, take the new closest result within the group (which could have changed since the time we updated the UI), and open the associated URL.

cco3:
I think a PwoGroup class would be nice, but I would isolate its use to the adapter, not to the whole fragment.

cco3:
There's some extra issues here in that sometimes we display urls before we've received the urlmetadata...we'll need some extra bin to store those in for when they receive their metadata...

nondebug:
I agree, it will make it much easier to work with grouped PWOs if we
introduce a PwoGroup abstraction. I think we'll also want a
time-independent abstraction for PWOs since it will be useful to track
information across successive PwoMetadatas from the same PWO.  That would
also help us solve the problem of matching up UrlMetadata when it comes
late.

nondebug:
I added a "NearbyPwoGroup" class inside the adapter to represent PWOs that share a group.  I also addded "NearbyPwo" to collect PwoMetadata that describe the same PWO, although currently it only keeps track of the most recent.

This doesn't work at all with the PROD endpoint because it currently ignores any PwoMetadata objects without groupid.  With the DEV endpoint PWOs won't appear until we have urlMetadata.  To fix this, we can add a bucket for group-less PWOs and promote them to grouped PWOs once we get urlMetadata.

nondebug:
Renamed NearbyPwo->PwoDevice and NearbyPwoGroup->PwoGroup.  PWOs without a group are once again displayed, which makes the experience much better when switching from PROD to DEV.  When we receive metadata for an ungrouped PWO that includes a groupid, the ungrouped PWO is promoted to a grouped PWO.

The display list can now contain a mix of PwoDevice and PwoGroup items.  A PwoDevice will expose the most recent metadata for that device, while a PwoGroup exposes the most recent metadata for the highest-ranking PWO in that group.

cco3:
LGTM

nuch-lek:
nuchlek22@gmail.com
เมื่อ 30 ก.ย. 2015 05:45 "Conley Owens" <notifications@github.com> เขียนว่า

PR #518, opened 2015-09-25
Use the URL shortener for non-ASCII URLs

nondebug:
Detect when a URL uses non-US-ASCII characters and automatically use the URL shortener.
If the shortener chokes on a URL, make sure we display an appropriate error message (ie, don't say the URL is too long if it isn't).


cco3:
Great!

PR #522, opened 2015-10-03
Add unit tests for BeaconDisplayList and fix bugs exposed by new tests

nondebug:
Two bugs fixed in this CL:

We were incorrectly merging two beacons advertising the same URL.  There are still cases where we will do this (eg, there's still no way to disambiguate non-BLE beacons advertising the same URL) but Bluetooth-based beacons can disambiguate by comparing device addresses.  With this CL, all beacons (grouped or ungrouped) use the same pwoId, which is "ble:"+deviceAddress for Bluetooth beacons and "url:"+requestUrl for other beacons.  Previously we used url for ungrouped beacons and deviceAddress for grouped beacons (with a fallback to url for non-Bluetooth).

The second bug was caused by failing to remove a beacon from a PwoGroup when its groupid changed.  We now remove the beacon from the group, as well as removing the group itself if removing the beacon causes it to be empty.

### DIFF
--- a/android/PhysicalWeb/app/src/androidTest/java/org/physical_web/physicalweb/BeaconDisplayListTest.java
+++ b/android/PhysicalWeb/app/src/androidTest/java/org/physical_web/physicalweb/BeaconDisplayListTest.java
@@ -1,0 +1,355 @@
+/*
+ * Copyright 2015 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


cco3:
LGTM

Thanks for setting up tests!

cco3:
Actually, could you also add the appropriate test command to travis/build.sh?

nondebug:
Sure, I'll look into how we can run these tests on Travis.

PR #524, opened 2015-10-05
Begin a Physical Web Collection library for java

cco3:
This java library contains data structures and convenience utilities
for storing metadata related to devices that broadcast URLs.  This
library is intended to help bootstrap new Physical Web clients written
in java.


java/libs/src/main/java/PwoCollection.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.physicalweb.collection;
+
+/**
+ * Collection of Physical Web Objects and related metadata.
+ */
+public class PwoCollection {
```
mmocny: Are we sure we want to stick with the "Pwo.." naming scheme?

cco3: Have a better suggestion?  I like it as long as we don't make a PwoDevice, etc.

PR #525, opened 2015-10-06
Create a SimplePwo class

cco3:
* Create a Pwo interface
* Create a SimplePwo class that just has an ID and a URL
* Add methods to add and fetch Pwos from PwoCollection


nondebug:
LGTM

PR #532, opened 2015-10-08
Add a UrlDeviceJsonSerializer to collection lib

cco3:
This class simplifies serialization a bit since it permits the
PhysicalWebCollection class to decide which serializer is associated
with which Pwo.


nondebug:
LGTM

PR #533, opened 2015-10-08
Add PwsMetadata to collection lib for storing URL metadata.

nondebug:
I'm using "requestUrl" to mean the URL broadcast by the beacon that is sent as a query to PWS, and "siteUrl" to mean the URL returned by PWS that we'll navigate to when the user selects the item.

### DIFF
--- a/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
+++ b/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
@@ -30,8 +30,12 @@ public class PhysicalWebCollection {
   private final String DEVICES_KEY = "devices";
   private final String TYPE_KEY = "type";
   private final String DATA_KEY = "data";
+  private final String METADATA_KEY = "metadata";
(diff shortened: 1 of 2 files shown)


nondebug:
Renamed to PwsResult, removed request->site url mapping

cco3:
LGTM
Optional: leave a serialization TODO

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -30,8 +30,12 @@
   private final String DEVICES_KEY = "devices";
   private final String TYPE_KEY = "type";
   private final String DATA_KEY = "data";
+  private final String METADATA_KEY = "metadata";
+  private final String REQUESTURL_KEY = "requesturl";
+  private final String SITEURL_KEY = "siteurl";
```
mmocny: "siteurl" is not currently used by PWS, nor has it been previously suggested going forward.

Would be worthwhile to share names as much as possible -- let's discuss tomorrow.

PR #536, opened 2015-10-08
Add metadata serialization and unit tests

nondebug:
This exposes "requestUrl" in a public method, any suggestions on a better name for this?  broadcastUrl maybe?

### DIFF
--- a/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
+++ b/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
@@ -35,7 +35,7 @@ public class PhysicalWebCollection {
   private static final String SITEURL_KEY = "siteurl";
   private Map<String, UrlDevice> mDeviceIdToUrlDeviceMap;
   private Map<Class, UrlDeviceJsonSerializer> mUrlDeviceTypeToUrlDeviceJsonSerializer;
-  private Map<String, PwsResult> mRequestUrlToPwsResultMap;
+  private Map<String, PwsResult> mBroadcastUrlToPwsResultMap;
 
   /**
    * Construct a PhysicalWebCollection.
@@ -43,7 +43,7 @@ public class PhysicalWebCollection {
   public PhysicalWebCollection() {
     mDeviceIdToUrlDeviceMap = new HashMap<>();
     mUrlDeviceTypeToUrlDeviceJsonSerializer = new HashMap<>();
-    mRequestUrlToPwsResultMap = new HashMap<>();
+    mBroadcastUrlToPwsResultMap = new HashMap<>();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


cco3:
Is deviceUrl too confusing?  I like broadcastUrl too.

nondebug:
Yeah, I think having both urlDevice and deviceUrl would be too confusing.  Let's use broadcastUrl.

cco3:
We could have UrlDeviceUrl...lol.

nondebug:
Renamed PWC public methods to use "broadcastUrl" instead of "requestUrl".

cco3:
LGTM

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -35,7 +35,7 @@
   private static final String SITEURL_KEY = "siteurl";
   private Map<String, UrlDevice> mDeviceIdToUrlDeviceMap;
   private Map<Class, UrlDeviceJsonSerializer> mUrlDeviceTypeToUrlDeviceJsonSerializer;
-  private Map<String, PwsResult> mRequestUrlToPwsResultMap;
+  private Map<String, PwsResult> mBroadcastUrlToPwsResultMap;
```
mmocny: My 2 cents: this is a "resolved url" to "url metadata" map, and we have a separate data structure for "broadcast url" to "resolved url".

The reason for this multiple beacons with different "broadcast urls" may resolve to the same PwsResult, and they should share.

And generally we may want to look up PwsResults for a specific url, without having to know which broadcast url was used to get the result..

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -72,6 +72,15 @@ public UrlDevice getUrlDeviceById(String id) {
   }
 
   /**
+   * Fetches cached URL metadata using the URL broadcasted by the Physical Web device.
+   * @param broadcastUrl The URL broadcasted by the device.
+   * @return Cached metadata relevant to the given URL.
+   */
+  public PwsResult getMetadataByBroadcastUrl(String broadcastUrl) {
+    return mBroadcastUrlToPwsResultMap.get(broadcastUrl);
```
mmocny: ...so this should look up resolved url for the given broadcast url and return the result for that.

If we don't have a mapping yet, we need to create one with a call to PWS.

We may also still not have one after call to PWS.

Should this return an async, given that it is the public interface?

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: I'm not convinced we gain much other than complexity from sharing data instead of duplicating it (and potentially swapping out old PwsResults with new ones when found).

Also, isn't it possible that the broadcast url could play into the the PwsResult in the future?  Couldn't myshadyshortener.net give a worse quality score compared to a private, trusted shortener?

Also, I'm not sure why you are saying that we generally want to look these entries up by their site url and not the broadcast url.  That isn't the case in the prototype app (nothing is ever keyed on the displayed url there).

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: Yea, I think a future would be good here.  That will make a good follow-up change.

PR #537, opened 2015-10-09
Add a PwPair class to the collection library

cco3:
A Physical Web pair is a UrlDevice and its corresponding UrlResult.
This change also provides an api for gathering these pairs from the
PhysicalWebCollection.

### DIFF
--- a/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
+++ b/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
@@ -18,8 +18,13 @@ package org.physical_web.collection;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Collection of Physical Web URL devices and related metadata.
@@ -210,5 +215,66 @@ public class PhysicalWebCollection {
       PwsResult pwsResult = jsonDeserializePwsResult(metadataPair);
       addMetadata(pwsResult);
     }
+  }
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order.
+   * @param dedupSiteUrls whether or not to dedup PwPairs by result urls.
+   * @param minRank the minimum rank to keep in the list.
+   * @return a sorted list of PwPairs.
+   */
+  public List<PwPair> getPwPairsSortedByRank(boolean dedupSiteUrls, double minRank) {
+    // Get all valid PwPairs.
+    List<PwPair> allPwPairs = new ArrayList<>();
+    for (UrlDevice urlDevice : mDeviceIdToUrlDeviceMap.values()) {
+      PwsResult pwsResult = mBroadcastUrlToPwsResultMap.get(urlDevice.getUrl());
+      if (pwsResult != null) {
+        allPwPairs.add(new PwPair(urlDevice, pwsResult));
+      }
+    }
+
+    // Sort the list in descending order.
+    Collections.sort(allPwPairs);
+    Collections.reverse(allPwPairs);
+
+    // Filter the list.
+    List<PwPair> ret = new ArrayList<>();
+    Set<String> siteUrls = new HashSet<>();
+    for (PwPair pwPair : allPwPairs) {
+      String siteUrl = pwPair.getPwsResult().getSiteUrl();
+      if (pwPair.getRank() >= minRank
+          && !dedupSiteUrls && !siteUrls.contains(siteUrl)) {
+        siteUrls.add(siteUrl);
+        ret.add(pwPair);
+      }
+    }
+
+    return ret;
+  }
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order.
+   * @param dedupSiteUrls whether or not to dedup PwPairs by result urls.
+   * @return a sorted list of PwPairs.
+   */
+  public List<PwPair> getPwPairsSortedByRank(boolean dedupSiteUrls) {
+    return getPwPairsSortedByRank(dedupSiteUrls, Double.NEGATIVE_INFINITY);
+  }
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order.
+   * @param minRank the minimum rank to keep in the list.
+   * @return a sorted list of PwPairs.
+   */
+  public List<PwPair> getPwPairsSortedByRank(double minRank) {
+    return getPwPairsSortedByRank(false, minRank);
+  }
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order.
+   * @return a sorted list of PwPairs.
+   */
+  public List<PwPair> getPwPairsSortedByRank() {
+    return getPwPairsSortedByRank(false, Double.NEGATIVE_INFINITY);
   }
 }
--- a/java/libs/src/main/java/org/physical_web/collection/PwPair.java
+++ b/java/libs/src/main/java/org/physical_web/collection/PwPair.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2015 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -211,4 +216,65 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
...
+    Collections.sort(allPwPairs);
+    Collections.reverse(allPwPairs);
+
+    // Filter the list.
+    List<PwPair> ret = new ArrayList<>();
+    Set<String> siteUrls = new HashSet<>();
+    for (PwPair pwPair : allPwPairs) {
+      String siteUrl = pwPair.getPwsResult().getSiteUrl();
+      if (pwPair.getRank() >= minRank
+          && !dedupSiteUrls && !siteUrls.contains(siteUrl)) {
```
mmocny: What if the PwsResults are not equal?  Could have different metadata/rank.  (e.g. in the case that the actual url content changed -- which comes up if you saw one beacon before the second one.  Probably not a very common problem).

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -211,4 +216,65 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
...
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order.
+   * @param dedupSiteUrls whether or not to dedup PwPairs by result urls.
+   * @param minRank the minimum rank to keep in the list.
+   * @return a sorted list of PwPairs.
+   */
+  public List<PwPair> getPwPairsSortedByRank(boolean dedupSiteUrls, double minRank) {
+    // Get all valid PwPairs.
+    List<PwPair> allPwPairs = new ArrayList<>();
```
mmocny: Why not build a Hash<siteUrl, PwPair> from the get go so that we don't have to do the second pass?

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -211,4 +216,65 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
...
+    for (UrlDevice urlDevice : mDeviceIdToUrlDeviceMap.values()) {
+      PwsResult pwsResult = mBroadcastUrlToPwsResultMap.get(urlDevice.getUrl());
+      if (pwsResult != null) {
+        allPwPairs.add(new PwPair(urlDevice, pwsResult));
+      }
+    }
+
+    // Sort the list in descending order.
+    Collections.sort(allPwPairs);
```
mmocny: Quick google finds: Collections.sort(list, Collections.reverseOrder());

I've not used this before so cannot vouch for it, but either way would rather create custom comparator than a separate reverse step.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -211,4 +216,65 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
...
     }
   }
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order.
+   * @param dedupSiteUrls whether or not to dedup PwPairs by result urls.
+   * @param minRank the minimum rank to keep in the list.
+   * @return a sorted list of PwPairs.
+   */
+  public List<PwPair> getPwPairsSortedByRank(boolean dedupSiteUrls, double minRank) {
```
mmocny: I do not understand the value of not deduping siteUrl.

cco3: Analytics

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: You mean the siteUrls hashset?  It wouldn't be valid if we constructed it here because there should be no entry before we've found the highest ranked pair.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: Really?  There's going to be...~50 of these things at most...doesn't seem like so much of a computation burden to warrant the code complexity.

Anyway, I've used your code.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: Then it gets the pair with the highest rank.

You are right that this is going to be a rare case...there would have to be multiple beacons broadcasting the same URL *and* we would have to make a request to PWS right as the cache expires.  I'm fine with "misbehavior" in that circumstance.  Anyway, if we want to replace the old PwsResult with a new one, that should happen at a separate step where we swap out the old PwsResult in mBroadcastUrlToPwsResultMap as soon as we find a new one.

cco3:
I'm going to submit this so we can start coding around it.  I'm happy to make changes after the fact.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
mmocny: I don't understand how that would be used.  You would just get the same results twice.

We could expose the list of <broadcastUrl, siteUrl> pairs, perhaps, for analytics?

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
mmocny: I agree, thats what we should do, swap out old PwsResult at resolve time.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: They wouldn't be the same results.  You'd get different UrlDevices, which gives you more information than just the broadcast url if you can cast it to your custom class.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -211,4 +216,65 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
...
+  public List<PwPair> getPwPairsSortedByRank(boolean dedupSiteUrls) {
+    return getPwPairsSortedByRank(dedupSiteUrls, Double.NEGATIVE_INFINITY);
+  }
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order.
+   * @param minRank the minimum rank to keep in the list.
+   * @return a sorted list of PwPairs.
+   */
+  public List<PwPair> getPwPairsSortedByRank(double minRank) {
```
mmocny: Imho, we should drop the minRank overloads.

The app can just follow up with a filter if it wants a minrank.  And I'm not sure we will necessarily be able to expose an actual rank number, we may only have an ordered list, so any meaning behind the actual values will fluctuate.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
mmocny: Why do we return `PwPair` here?  I think we should just return `PwsResult` list.

Returning the UrlDevice may have uses for debugging, but we conceivably may want to return PwsResult's which come from sources other than UrlDevices.

mmocny: Okay, if we do want to return UrlDevices, how about we rename the existing `PwsResult` to `PageMetadata` (or something), and change `getPwPairsSortedByRank` to return a `List<PwResult>` where `PwResult` is:

```
class PwResult {
   PageMetadata page_md;
   List<UrlDevice> devices;
}
```

Then we can always have deduplicated results, and still have access to all the beacon info if we want it.

WDYT?

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: The rank is implemented by custom UrlDevice definitions, so even in the case of an ordered list, one would be created.

We can drop the overloads though

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
nondebug: I like this proposal.  Does PwResult.devices contain all UrlDevices for a particular broadcastUrl or siteUrl?  I would lean toward siteUrl since grouping by broadcastUrl isn't as useful.

mmocny: Yes, based on siteUrl.  We could even sort the devices by relative order.

PR #549, opened 2015-10-13
Add makeJavadoc task and fix javadoc errors

nondebug:


cco3:
Does this ensure every public method/class has javadoc?

cco3:
LGTM

nondebug:
It doesn't enforce javadoc on public members, but that would be a good idea.  Currently it will only warn if you've already created a javadoc block.

nondebug:
Investigation into warning on missing javadoc indicates that we can pass this javac flag to enable doclint when the library is built:

# enable doclint for all public and protected members
-Xdoclint:all/protected

However I couldn't find a way to disable this for tests, so I left it out.  We're currently clean for all non-test java files.

I added a few more non-javadoc fixes to get a clean build:

* Exclude findbugs warning for an anonymous inner class that could be made a static class.
* Remove trailing whitespace in IntegrationTest.java

PR #551, opened 2015-10-13
Add URL group IDs in PwsResult

nondebug:
Add groupId field in PwsResult (serialization treats it as optional)
Add UrlGroup container for returning grouped device/metadata pairs
Add getUrlGroupsSortedByRank (groups are compared/sorted based on their top-ranked PwPairs)
Move RankedDevice mock object out of PhysicalWebCollectionTest, it is also used by UrlGroupTest

### DIFF
--- a/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
+++ b/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
@@ -39,6 +39,7 @@ public class PhysicalWebCollection {
   private static final String METADATA_KEY = "metadata";
   private static final String REQUESTURL_KEY = "requesturl";
   private static final String SITEURL_KEY = "siteurl";
+  private static final String GROUPID_KEY = "groupid";
   private static final String DEFAULT_PWS_ENDPOINT = "https://url-caster.appspot.com";
   private PwsClient mPwsClient;
   private Map<String, UrlDevice> mDeviceIdToUrlDeviceMap;
@@ -130,6 +131,12 @@ public class PhysicalWebCollection {
     JSONObject jsonObject = new JSONObject();
     jsonObject.put(REQUESTURL_KEY, pwsResult.getRequestUrl());
     jsonObject.put(SITEURL_KEY, pwsResult.getSiteUrl());
+
+    String groupId = pwsResult.getGroupId();
+    if (groupId != null && !groupId.equals("")) {
+      jsonObject.put(GROUPID_KEY, groupId);
+    }
+
     return jsonObject;
   }
 
@@ -192,7 +199,11 @@ public class PhysicalWebCollection {
   private PwsResult jsonDeserializePwsResult(JSONObject jsonObject) {
     String requestUrl = jsonObject.getString(REQUESTURL_KEY);
     String siteUrl = jsonObject.getString(SITEURL_KEY);
-    return new PwsResult(requestUrl, siteUrl);
+    String groupId = null;
+    if (jsonObject.has(GROUPID_KEY)) {
+      groupId = jsonObject.getString(GROUPID_KEY);
+    }
+    return new PwsResult(requestUrl, siteUrl, groupId);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


nondebug:
Didn't realize pushing a new version would delete all the comments... oops

> cco3: Why do we return a list of groups and now just a filtered down set of PwPairs?
> cco3: Hm, thinking about this a bit more and I see how it could be useful, but I'm not sure what the best way to expose this is.

I think it would be more useful to return PwPairs since in most cases the user is just going to display them in a list.  If the user needs to find all PwPairs with the same groupid they can call getPwPairsSortedByRank and filter it themselves.

nondebug:
> cco3: is this test necessary? We are always going to have a groupid unless something has gone wrong, right? And if we don't have a groupid, it doesn't matter how we serialize it, does it?

It's probably fine to serialize null and missing groupids the same way since they currently have the same behavior, but I didn't want to make any assumptions.  Once groupid is included for all PWS results we can make it mandatory and remove the extra checks.

cco3:
LGTM

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -260,6 +271,38 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
...
+   * Return a list of UrlGroups sorted by the rank of the top-ranked pair in each group.
+   * @return a sorted list of UrlGroups.
+   */
+  public List<UrlGroup> getUrlGroupsSortedByRank() {
+    List<PwPair> sortedPwPairs = getPwPairsSortedByRank();
+
+    Map<String, UrlGroup> pwGroupMap = new HashMap<>();
+    for (PwPair pwPair : sortedPwPairs) {
+      PwsResult pwsResult = pwPair.getPwsResult();
+      if (pwsResult != null) {
```
mmocny: Can this be null?

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -260,6 +271,38 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
...
+        String groupId = pwsResult.getGroupId();
+        if (groupId != null && !groupId.equals("")) {
```
mmocny: Does this mean that you do not include results which have no group?  Shouldn't we still include them?

java/libs/src/main/java/org/physical_web/collection/UrlGroup.java
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Check if two UrlGroups are equal based on the ranks of their top pairs.
+   * @param other the UrlGroup to compare to.
+   * @return true if the top pairs are of equal rank.
+   */
+  @Override
+  public boolean equals(Object other) {
+    if (other instanceof UrlGroup) {
+      UrlGroup otherUrlGroup = (UrlGroup) other;
+      return getTopPair() == otherUrlGroup.getTopPair();
```
mmocny: Is this enough for equality?  Seems like it would bite us at some point.  Why not just `return mPwPairs.equals(((UrlGroup) other).mPwPairs)`?

mmocny: ..`&& mGroupId == other.mGroupId` probably.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: no, it can't...we should remove that check

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
cco3: Assuming PWS hasn't left them out for some reason, no...we should remove this check too.

java/libs/src/main/java/org/physical_web/collection/UrlGroup.java
cco3: +1 on this too.

nondebug: Making equals more strict would cause it to be inconsistent with compareTo:

> The natural ordering for a class C is said to be consistent with equals if and only if e1.compareTo(e2) == 0 has the same boolean value as e1.equals(e2) for every e1 and e2 of class C. Note that null is not an instance of any class, and e.compareTo(null) should throw a NullPointerException even though e.equals(null) returns false.

> It is strongly recommended (though not required) that natural orderings be consistent with equals. This is so because sorted sets (and sorted maps) without explicit comparators behave "strangely" when they are used with elements (or keys) whose natural ordering is inconsistent with equals. In particular, such a sorted set (or sorted map) violates the general contract for set (or map), which is defined in terms of the equals method.

http://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
nondebug: Correct, we're not including ungrouped results.  It's "getUrlGroupsSortedByRank" so this is by design.  I agree a "get displayable PwPairs" method that sorts and returns the top group results combined with ungrouped results would be more useful.

Currently it's still valid for PWS to return no groupid.  In the future we'll require groupid, correct?  If so, we should validate when we deserialize PwsResult rather than here.  On the other hand if we decide null/empty groupid is valid then we'll still need to filter here.

mmocny: I think we can assume empty is valid (though current PWS will always return a result) and still get the api that we want.

Just create an empty group for each PwPair that doesn't have a groupid.

I think this api is meant to be: get me all valid results, but group them.

mmocny: ..clarification: current PWS codebase will always return a result but the feature has not been deployed to stable.  Can do that once we are comfortable (I'm happy to do it asap but I don't have time to test right now, if anyone wants to run DEV through the test gauntlet just let me know).

java/libs/src/main/java/org/physical_web/collection/UrlGroup.java
mmocny: We want comparable to compare lists according to closest->furthest but equal to compare everything..

Perhaps we just maintain mPwPairs in sorted order?  I expect we will modify the list less frequently than we would access `getTopPair()` so I don't actually mind swapping O(1) add + O(n) lookup for O(logn) add + O(1) lookup.

nondebug: I have an open PR (https://github.com/google/physical-web/pull/558) with the suggested changes from this PR, let's move discussion there.

PR #553, opened 2015-10-14
Add location permission

cco3:
This allows to do scanning on screen wakeup on M.


cco3:
FYI @nondebug

cco3:
new change up with coarse location...apparently that's good enough

g-ortuno:
LGTM

PR #557, opened 2015-10-15
Fix #556 - Persist url fragment in PWS fetch

### LINKED ISSUE #556: Anchor tags are stripped from URLs
Configure a beacon to have a URL with a trailing anchor tag, e.g. 'http://cnn.com#test'

Find the beacon in the Physical Web app, you'll see CNN but notice that the URL looses the tag

Just to be sure, click on the URL and you'll notice that you go to cnn.com

mmocny:


cco3:
LGTM

mmocny:
Going to push this change to DEV so we can try it live.

PR #558, opened 2015-10-15
Clean up groupid functionality in physical web library

nondebug:
Remove unnecessary null check on pwsResult in getUrlGroupsSortedByRank
Break ties when comparing UrlGroups by comparing groupid strings


cco3:
LGTM

PR #559, opened 2015-10-16
Modify API for retrieving grouped URLs

nondebug:
Return a list of PwPairs instead of UrlGroups
Include ungrouped URLs in the returned list

### DIFF
--- a/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
+++ b/java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
@@ -245,6 +245,39 @@ public class PhysicalWebCollection {
    */
   public List<PwPair> getPwPairsSortedByRank() {
     // Get all valid PwPairs.
+    List<PwPair> allPwPairs = getPwPairs();
+
+    // Sort the list in descending order.
+    Collections.sort(allPwPairs, Collections.reverseOrder());
+
+    // Filter the list.
+    return removeDuplicateSiteUrls(allPwPairs);
+  }
+
+  /**
+   * Return a list of PwPairs sorted by rank in descending order, including only the top-ranked
+   * pair from each group.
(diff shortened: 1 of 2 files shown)


java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -245,6 +245,39 @@ public void jsonDeserialize(JSONObject jsonObject) throws PhysicalWebCollectionE
    */
   public List<PwPair> getPwPairsSortedByRank() {
     // Get all valid PwPairs.
+    List<PwPair> allPwPairs = getPwPairs();
```
dinhviethoa: Maybe I'm late here but why returning a list of PwPair and not a simply list or PwResult.
The PwResult could embed the UrlDevice or could include at least the initial URL of the UrlDevice.
That would be one less datatype to deal with from the API perspective.

nondebug: In our abstraction UrlDevice represents the beacon and PwsResult represents the cached URL metadata from PWS.  The problem with attaching UrlDevice to PwsResult is that multiple devices could be broadcasting the same URL, and we only keep one PwsResult per broadcast URL.  Instead of one UrlDevice reference the PwsResult would need to keep a list and pick the correct one based on rank.  Maintaining this list as beacons are found and lost is likely more complex than building the associations and filtering them at display time as we do now.

(btw, we actually do store the initial URL of the UrlDevice in PwsResult as mRequestUrl.)

I think there's value in keeping the PWS-derived metadata and beacon-derived metadata logically separated.  PwPair does serve a purpose in providing a place to attach behaviors (like ordering) that depend on both the device and URL metadata.

cco3:
Other than the 2 naming comments, LGTM

PR #565, opened 2015-10-26
Make http classes package private

cco3:
There are several classe that don't need to be exposed to developers.
This change makes them package private.


nondebug:
LGTM

PR #586, opened 2016-01-25
Allow client code to fetch images from collection

cco3:
This change allows client code to fetch images directly from the
Physical Web Collection by supplying a callback.


nondebug:
lgtm

PR #587, opened 2016-01-27
Make UrlDevice a class

cco3:
UrlDevice was an interface, the idea being that we wanted to encourage
users to implement this themselves to add extra data.  Instead, we
should have a class that allows users to add extra data to it.  A
follow up change will let user supply key/value pairs.


cco3:
This was about as small as I could make it, sorry it's still so large.

PR #590, opened 2016-02-11
Add extra data to PwsResult and UrlDevice

cco3:
Add putters and getters for PwsResult and UrlDevice so that clients
can store arbitrary data associated with each.  This is better than
our previous solution of requiring client code to subclass these.


cco3:
Updated

cco3:
Updated

nondebug:
lgtm

PR #591, opened 2016-02-16
Share data directly between Fragment and Service

cco3:
Instead of subscribing to every detailed update, the fragment now
shares data directly with the service and receives notices when
any data has changed.  This simplifies the interactions between the
two, paving the way for a shared Collection (when we integrate the
collections library).

One aspect of this change is that it replaces the fragments
mPwoMetadataQueue with an mUrlQueue.  This is simplier for checking
whether new urls are already stored in the queue, but it also will
allow us to access the most appropriate metadata at the time of
display, as opposed to at the time of placing the data into the
queue.


nondebug:
lgtm

PR #614, opened 2016-03-15
Responsive embeds and heading line height

jaicab:
The YouTube embeds on the website weren't responsive, causing a increase in layout width on mobile, breaking the experience. Using the common pattern of 16:9 aspect ratio for responsive embeds via CSS I applied it to all the iframes in the HTML.

I also fixed a line height problem in the headings as it was being relied on for margin. In cases where the headings go to two lines, the spacing was huge. So changed it to 1 and applied margin bottom.


jaicab:
Oops! I compared it with the wrong branch. Fixed on #615.

PR #616, opened 2016-03-16
[web-service] Add secureOnly options to body request

beaufortfrancois:
R=@mmocny

### DIFF
--- a/web-service/handlers.py
+++ b/web-service/handlers.py
@@ -72,10 +72,12 @@ class ResolveScan(webapp2.RequestHandler):
         try:
             input_object = json.loads(input_data) # TODO: Data is not sanitised.
             objects = input_object.get('objects', [])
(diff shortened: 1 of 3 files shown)


mmocny:
Francois, this LGTM.  From the looks of it the HTTPS requirement is only on the final URL which is what we want.  And thanks for some adding tests!

However, I'm off on vacation until after Easter and even if I merge, will not be able to deploy.

@cco3 and @nondebug feel free to merge.

oahziur:
I might be able to help deploy this if necessary. I think I have the access to both `url-caster-dev` and `url-caster` App Engine project. Otherwise, we can wait after Easter.

beaufortfrancois:
Please let me know when it's live.

oahziur:
@beaufortfrancois Both `url-caster` and `url-caster-dev` are up to date.

beaufortfrancois:
Thank you!

PR #617, opened 2016-03-17
Add optExtra methods to PwsResult and UrlDevice

cco3:


java/libs/src/main/java/org/physical_web/collection/UrlDevice.java
```diff
@@ -159,49 +159,153 @@ public String getUrl() {
...
     return mExtraData.optJSONArray(key);
   }
 
   /**
    * Get extra JSONObject value.
-   * @return The stored value or null if it doesn't exist in specified form.
+   * @param key The key of the stored value.
+   * @return The stored value.
    */
   public JSONObject getExtraJSONObject(String key) {
```
nondebug: addExtra(String, Object) can add "Any object of type JSONObject, JSONArray, String, Boolean, Integer, Long, Double, NULL, or null. May not be NaNs or infinities" but currently we have no way to retrieve a string value.

java/libs/src/main/java/org/physical_web/collection/PwsResult.java
```diff
@@ -302,49 +302,153 @@ public String getGroupId() {
 
   /**
    * Get extra boolean value.
+   * @param key The key of the stored value.
    * @return The stored value or null if it doesn't exist in specified form.
```
nondebug: The getExtra.* methods should say we'll throw JSONException if the key isn't present

cco3:
Updated

nondebug:
lgtm

PR #618, opened 2016-03-17
Integrate new library into PW android app

cco3:
Behavior change: we no longer show URLs that have not resolved.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -363,34 +371,55 @@ public void onAnimationCancel(Animator animation) {
...
 
-    public void addItem(PwoMetadata pwoMetadata) {
-      mBeaconDisplayList.addItem(pwoMetadata);
+    public void addItem(PwPair pwPair) {
+      mPwPairs.add(pwPair);
     }
 
-    public boolean containsUrl(String url) {
-      return mBeaconDisplayList.containsUrl(url);
+    public void updateItem(PwPair pwPair) {
```
nondebug: Would it make sense to combine addItem and updateItem into one method that does the right thing regardless of whether the item is already in the collection?

cco3: They are indexed in an arraylist, so add would become a linear operation instead of constant.

nondebug:
lgtm

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconDisplayList.java
```diff
@@ -1,309 +0,0 @@
...
-
-    /**
-     * Append new metadata for this PWO.
-     * @param pwoMetadata Updated metadata for this PWO
-     */
-    public void addItem(PwoMetadata pwoMetadata) {
-      // TODO(mattreynolds): track PWO history (for now, just keep the most recent)
-      mPwoMetadata = pwoMetadata;
-    }
-
```
TechVinothkumar: adssdasdadsdadad

Issue #654, opened 2016-05-13
Resolve Scan returning an error


kevinahuber:
Two URLs: 
```
https://www.cvs.com/deals/deals.jsp
https://www.kroger.com/topic/in-store-services
```
are not working with the physical web resolve service, http://service.physical-web.org/resolve-scan, but are valid URLs. Here is my error message:
```
Traceback (most recent call last):
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.5.2/webapp2.py", line 1535, in __call__
    rv = self.handle_exception(request, response, e)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.5.2/webapp2.py", line 1529, in __call__
    rv = self.router.dispatch(request, response)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.5.2/webapp2.py", line 1278, in default_dispatcher
    return route.handler_adapter(request, response)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.5.2/webapp2.py", line 1102, in __call__
    return handler.dispatch()
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.5.2/webapp2.py", line 572, in dispatch
    return self.handle_exception(e, self.app.debug)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.5.2/webapp2.py", line 570, in dispatch
    return method(*args, **kwargs)
  File "/base/data/home/apps/s~url-caster/1.391444449619714950/handlers.py", line 80, in post
    output = helpers.BuildResponse(objects, secure_only)
  File "/base/data/home/apps/s~url-caster/1.391444449619714950/helpers.py", line 106, in BuildResponse
    logging.error('ComputeGroupId url:{0}, title:{1}, description:{2}'.format(siteInfo.url, siteInfo.title, siteInfo.description))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 86: ordinal not in range(128)
```

It looks like `\xa0` is not handled well. We could use `unicodedata.normalize()` before sending the results to prevent this and future errors. I cannot send in a fix now, but will be able to this weekend.

scottjenson:
@kevinahuber I'm cleaning up old Issues that seem to have gone stale. Is this still an issue? If so, please reopen and we can discuss.

kevinahuber:
[...]

PR #655, opened 2016-05-15
Group ID Formation Fix

kevinahuber:
Fixes the issue caused by the first URL in #654: `https://www.cvs.com/deals/deals.jsp`

When a group id is formed, it fails if the title is `None`. Rather than just using the domain when the title is `None`, I am proposing that we fall back to the description, then to the path. This would ensure that the group ID would fall back to be more unique, rather than less.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -35,3 +35,4 @@ louaybassbouss
 devunwired
 ColeMurray
 edent
+kevinahuber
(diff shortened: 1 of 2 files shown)


kevinahuber:
I signed it!

kevinahuber:
...messed up my CLA with another email! Will open a new one.

PR #656, opened 2016-05-15
Group ID Formation Fix

kevinahuber:
Fixes the issue caused by the first URL in #654: https://www.cvs.com/deals/deals.jsp

When a group id is formed, it fails if the title is None. Rather than just using the domain when the title is None, I am proposing that we fall back to the description, then to the path. This would ensure that the group ID would fall back to be more unique, rather than less.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -35,3 +35,4 @@ louaybassbouss
 devunwired
 ColeMurray
 edent
+kevinahuber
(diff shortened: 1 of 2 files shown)


cco3:
LGTM, assigning to Rui.

kevinahuber:
@oahziur Doubling back on this, hows it look?

cco3:
Adding @mmocny also.

oahziur:
LGTM

Issue #657, opened 2016-05-16
Document MDNS as non-working


raggi:
As far as I can see MDNS isn't working today. Can this get documented, as I tried and lost time trying.

cco3:
Right, we disabled it due to several issues.  I didn't notice that we didn't note that in the documentation.  thanks!

scottjenson:
It's now working on Android M. There is a lot more to do but it's a start.

PR #661, opened 2016-05-19
Add note regarding current disabled status of mDNS

cco3:


cco3:
#657

documentation/mDNS_Support.md
```diff
@@ -1,5 +1,7 @@
 ## mDNS Support
 
+**Due to several issues, mDNS support is currently disabled this app!  We intend to revisit (and hopefully reenable) this feature in the future.**
```
nondebug: disabled *in this app

IMO "re-enable" reads better than "reenable"

raggi:
I'm also wondering if this is true of the android chrome implementation. I didn't have an mdns server to hand that would allow me to generate the structure that avahi would, as the one I had was too strict.

Across the board, it'd be nice to know what is and isn't supported by the chrome implementation, especially as the chrome implementation links to this repo in docs. It appears by experimentation that ssdp and bluetooth work in chrome the way they are described here.

cco3:
The android chrome implementation doesn't have mdns support.  This is a test bed and will typically have always have features not in the chrome implementation.  ssdp also doesn't work in chrome...not sure why it appears to for you.

I'm not sure the docs in this repo are the best place to describe the feature sets of other PW clients.  It's something we can consider though.

raggi:
e.g. https://support.google.com/chrome/answer/6239299 links to this repo.

I thought I saw something from developers.google.com too, but I may have got my wires crossed.

PR #664, opened 2016-05-24
Fixed crash caused by modifying the view on a thread other than UI. F…

hayesjordan:
Fix crash due to invalid thread activity.


cco3:
Looks good, just some comments on the commit message:
* Use imperative mood ("Fix" instead of "Fixed")
* Limit the subject line to 50 characters
* Limit the message body to 72 characters

To edit your commit message, you can do `git commit -a --amend`

hayesjordan:
Updated the commit message.

PR #666, opened 2016-05-24
Update message when beacon is found while editing

hayesjordan:


android/PhysicalWeb/app/src/main/res/values/strings.xml
```diff
@@ -11,7 +11,7 @@
     <string name="nearby_beacon_favicon_content_description">favicon for the given URL for this
         entry</string>
     <string name="config_searching_for_beacons_text">Put your beacon into config mode</string>
-    <string name="config_found_near_beacon">Beacon found!</string>
+    <string name="config_found_near_beacon">Beacon found! Connecting to beacon.</string>
```
cco3: @scottjenson Do you have any suggestions for text here?

PR #668, opened 2016-05-26
Add HTTPS only to Android only

beaufortfrancois:
Hopefully we'll mark HTTPS only for iOS soon but for now we should be honest about it with developers and communicate this clearly.


cco3:
Thanks!

PR #673, opened 2016-05-31
Add Google's PWS

hayesjordan:


java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -61,8 +78,42 @@ public void setPwsEndpoint(String pwsEndpoint) {
...
+   */
+  public void setPwsApiKey(String pwsApiKey){
+      apiKey = pwsApiKey;
+  }
+
+  /**
+   * Set the API version the PWS endpoint uses.
+   * @param pwsApiVersion The new PWS endpoint.
+   */
+  public void setPwsApiVersion(int pwsApiVersion){
```
cco3: I think it'd be better if we didn't have this method...just add a second param to setEndpoint.

java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -61,8 +78,42 @@ public void setPwsEndpoint(String pwsEndpoint) {
     mPwsEndpoint = pwsEndpoint;
   }
 
-  private String constructPwsUrl(String path) {
-    return mPwsEndpoint + "/" + path;
+  /**
+   * Set the API key the PWS endpoint uses.
+   * @param pwsApiKey The new PWS endpoint.
+   */
+  public void setPwsApiKey(String pwsApiKey){
```
cco3: Same here as below.

java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -61,8 +78,42 @@ public void setPwsEndpoint(String pwsEndpoint) {
...
+  }
+
+  private String constructPwsResolveUrl() {
+    switch(apiVersion){
+      case 1:
+        return mPwsEndpoint + "/" + V1_RESOLVE_SCAN_PATH;
+      case 2:
+       return mPwsEndpoint + "/" + V2_RESOLVE_SCAN_PATH + "?key=" + apiKey;
+      default:
+        return "";
```
cco3: raise RuntimeException("Unknown API version").

java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -61,8 +78,42 @@ public void setPwsEndpoint(String pwsEndpoint) {
...
+    }
+  }
+
+  private String getPayloadApiJsonWrapper(){
```
cco3: Style nit: you should have spaces after the params.  I think `./gradlew check` will point out style issues.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -298,6 +298,22 @@ public void setPwsEndpoint(String pwsEndpoint) {
     mPwsClient.setPwsEndpoint(pwsEndpoint);
   }
 
+  /**
+   * Set the API version the PWS endpoint uses.
+   * @param pwsApiVersion The new PWS endpoint.
+   */
+  public void setPwsApiVersion(int pwsApiVersion) {
```
cco3: Same comments here as in the PwsClient...I think it would be good just to make these params to one method.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -54,6 +58,19 @@ private static void throwEncodeException(JSONException e) {
     throw new RuntimeException("Could not encode JSON", e);
   }
 
+  public static String getApiKey(Context context){
+    int resourceId = context.getResources().getIdentifier("google_api_key", "string", context.getPackageName());
+    return resourceId != 0 ? context.getString(resourceId) : "";
```
cco3: just do context.getString(R.string.google_api_key)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -54,6 +58,19 @@ private static void throwEncodeException(JSONException e) {
...
+  }
+
+  public static String getProdPwsEndpoint(String key){
```
cco3: These methods are a bit confusing in that it's weird that we would decide which endpoint to use based on what the key is.  Instead, let's make these all one method where you would pass a context and the pwcollection and set the values directly on the pwcollection.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
hayesjordan: The way I thought this was going to be used, the google_api_key wasn't going to be present instead of being there but empty. When its the resource doesn't exist context.getString(R.string.google_api_key) causes an compile error.

Do you want me to change it, and always have a string resource named google_api_key?

cco3: Ah, I didn't consider that it wouldn't compile.  Makes sense.

hayesjordan:
Updated per comments.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -454,14 +454,15 @@ public View getView(int i, View view, ViewGroup viewGroup) {
         updateDebugView(pwPair, view);
         view.findViewById(R.id.ranging_debug_container).setVisibility(View.VISIBLE);
         view.findViewById(R.id.metadata_debug_container).setVisibility(View.VISIBLE);
-        mPwCollection.setPwsEndpoint(Utils.DEV_ENDPOINT);
+        mPwCollection.setPwsEndpoint(Utils.DEV_ENDPOINT, Utils.DEV_ENDPOINT_VERSION, "");
```
cco3: Please make a version of this method that just takes two parameters that calls the three parameter method.  Also, I think `null` would be a better empty key rather than an empty string.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -454,14 +454,15 @@ public View getView(int i, View view, ViewGroup viewGroup) {
...
         UrlShortenerClient.getInstance(getActivity()).setEndpoint(Utils.DEV_ENDPOINT);
       } else {
         // Otherwise ensure it is not shown
         view.findViewById(R.id.ranging_debug_container).setVisibility(View.GONE);
         view.findViewById(R.id.metadata_debug_container).setVisibility(View.GONE);
-        mPwCollection.setPwsEndpoint(Utils.PROD_ENDPOINT);
-        UrlShortenerClient.getInstance(getActivity()).setEndpoint(Utils.PROD_ENDPOINT);
+        Utils.setPwsEndpoint(view.getContext(), mPwCollection);
```
cco3: Use getActivity() instead of view.getContext()

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -454,14 +454,15 @@ public View getView(int i, View view, ViewGroup viewGroup) {
...
+        UrlShortenerClient.getInstance(getActivity())
+          .setEndpoint(mPwCollection.getPwsEndpointUrl());
```
cco3: This isn't always going to work.  Only the open source Pws does shortening.  You should leave the line as it was, pointing to PROD_ENDPOINT

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -54,6 +59,34 @@ private static void throwEncodeException(JSONException e) {
     throw new RuntimeException("Could not encode JSON", e);
   }
 
+  private static String getApiKey(Context context){
+    int resourceId = context.getResources().getIdentifier("google_api_key", "string",
+                                                          context.getPackageName());
+    return resourceId != 0 ? context.getString(resourceId) : "";
+  }
+
+  public static void setPwsEndpoint(Context context, PwsClient pwsClient){
```
cco3: Given that this picks between two specific endpoints (and not the dev endpoint or other endpoints), maybe this should be renamed `setProdPwsEndpoint`.

java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -33,36 +33,78 @@
...
 
   /**
-   * Set the URL for making PWS requests.
+   * Set the URL, the API version, the API Key for making PWS requests.
    * @param pwsEndpoint The new PWS endpoint.
+   * @param pwsApiVersion The new PWS API version.
+   * @param pwsApiKey The new PWS API key.
    */
-  public void setPwsEndpoint(String pwsEndpoint) {
+  public void setPwsEndpoint(String pwsEndpoint, int pwsApiVersion, String pwsApiKey) {
```
cco3: I think this should be renamed to just `setEndpoint`.

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -291,11 +291,20 @@ public PwPair getTopRankedPwPairByGroupId(String groupId) {
...
-  public void setPwsEndpoint(String pwsEndpoint) {
-    mPwsClient.setPwsEndpoint(pwsEndpoint);
+  public void setPwsEndpoint(String pwsEndpoint, int pwsApiVersion, String pwsApiKey) {
+    mPwsClient.setPwsEndpoint(pwsEndpoint, pwsApiVersion, pwsApiKey);
+  }
+
+  /**
+   * Get the URL that is used to make PWS requests.
+   */
+  public String getPwsEndpointUrl(){
```
cco3: This won't be needed after you make the fix I suggested where you use it.

java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -33,36 +33,78 @@
...
     mPwsEndpoint = pwsEndpoint;
+    apiVersion = pwsApiVersion;
+    apiKey = pwsApiKey;
```
cco3: test if apiKey is null or empty and version is >=2. If so, throw a RuntimeException

java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -33,36 +33,78 @@
...
+      case 1:
+        return mPwsEndpoint + "/" + V1_RESOLVE_SCAN_PATH;
+      case 2:
+       return mPwsEndpoint + "/" + V2_RESOLVE_SCAN_PATH + "?key=" + apiKey;
+      default:
+        throw new RuntimeException(UKNOWN_API_ERROR_MESSAGE);
+    }
+  }
+
+  private String getPayloadApiJsonWrapper(){
```
cco3: I think this code should go directly in the method where you use it.  Breaking this out into its own method sort of suggests that the format of the payload is always the same, it's just the key that changes, which is kind of true right now, but not really, and probably won't be the case going forward.

java/libs/src/main/java/org/physical_web/collection/PwsClient.java
```diff
@@ -79,13 +121,67 @@ private void recordResponse() {
...
+      private PwsResult getV2PwsResult(JSONObject jsonUrlMetadata){
+        try {
+            JSONObject jsonPageInfo = jsonUrlMetadata.getJSONObject("pageInfo");
+            return new PwsResult.Builder(
+                jsonUrlMetadata.getString("scannedUrl"), jsonUrlMetadata.getString("resolvedUrl"))
+                .setTitle(jsonPageInfo.optString("title"))
+                .setDescription(jsonPageInfo.optString("description"))
+                .setIconUrl(jsonPageInfo.optString("icon"))
+                // TODO(hayesjordan): Add group id when V2 is changed to have group id.
+                .setGroupId("")
```
cco3: can you leave this unset and correct any errors that get caused by it being null?

cco3:
Looking good!  Just a few more things to cleanup.  Also, you can run `./gradlew check` to simulate the travis build.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
hayesjordan: This will be changed once the settings get added, so I am inclined to leave the generic name.

hayesjordan:
Updated per comments. Also I ran `./gradlew check` and it passes.

cco3:
LGTM!

PR #674, opened 2016-05-31
Add Share funtionality

nishadg246:
Allows users to share URLs through the native Android share menu. Doing
so will effectively allow your device, just as Physical Web beacons do,
to broadcast the URL you wish to share.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.List;
+import java.util.regex.Matcher;
+
+public class BroadcastActivity extends Activity {
+
+    private static final String TAG  ="BroadcastActivity";
+    private static final int ENABLE_BLUETOOTH_REQUEST_ID = 1;
+    public static final int ACCOUNT_PICKER_REQUEST_ID = 2;
+    public static final int PERMISSIONS_REQUEST_ID = 3;
+    public static final int GOOGLE_PLAY_SERVICES_REQUEST_ID = 5;
```
cco3: A lot of these don't look needed.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+        // Make sure the request was successful
+        Log.d(TAG,resultCode+"");
+        if(resultCode==-1){
+            if (!checkIfRequirement2Fulfilled()) {
+                return;
+            }
+            preBroadcast();
+            finish();
+        }else{
+            Toast.makeText(this, "Turn Bluetooth on to broadcast", Toast.LENGTH_LONG).show();
```
cco3: Don't we have a better way of doing this right now? (that actually prompts the user and then starts broadcasting when they have accepted?)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+            finish();
+        }
+        
+      }  // Do something with the contact here (bigger example below)
+    }
+      
+
+    @Override
+    protected void onStart() {
+        Log.d(TAG, "onStart");
```
cco3: This isn't needed

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+
+        if (!checkIfBluetoothCanAdvertise()) {
+            Toast.makeText(this, "Bluetooth LE advertising is not supported on this device", Toast.LENGTH_LONG).show();
+            finishAndRemoveTask();
+            return false;
+        }
+
+        return true;
+    }
+    private boolean checkIfRequirement2Fulfilled() {
```
cco3: What's the difference?  Could you make these method names more meaningful?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private void preBroadcast(){
+        Intent intent = getIntent();
+        String action = intent.getAction();
+        String type = intent.getType();
+        if (action.equals(Intent.ACTION_SEND) && (type != null)) {
+            Log.d(TAG, "content found!");
+            if (type.equals("text/plain")) {
+                Log.d(TAG, "text received");
```
cco3: Remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+                String text = intent.getStringExtra(Intent.EXTRA_TEXT);
+                // Look to see if the text is a URL
+                ourl = findUrlInText(text);
```
cco3: Just `url` might be better

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+        if (bluetoothManager.getAdapter().getBluetoothLeAdvertiser() == null) {
+            Log.d(TAG, "cant advertise");
+            return false;
+        }
+        Log.d(TAG, "can advertise");
+        return true;
+    }
+
+    private void setter(String url){
+        Intent inten = new Intent(this, HereYouGoService.class);
```
cco3: I think you use `inten` in several places.  `intent` weould be better.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,242 @@
...
+        BluetoothManager bluetoothManager = (BluetoothManager) getSystemService(Context.BLUETOOTH_SERVICE);
+        if (bluetoothManager.getAdapter().getBluetoothLeAdvertiser() == null) {
+            Log.d(TAG, "cant advertise");
+            return false;
+        }
+        Log.d(TAG, "can advertise");
+        return true;
+    }
+
+    private void setter(String url){
```
cco3: You could rename this to `startBroadcastService`

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/ConfigUriBeacon.java
```diff
@@ -0,0 +1,287 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.net.URISyntaxException;
+
+public class ConfigUriBeacon extends UriBeacon {
```
cco3: We don't need this whole class...just pull out the one utility method you use in throw it in one of the utility classes we have.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+import android.os.ParcelUuid;
+import android.preference.PreferenceManager;
+import android.support.v4.app.NotificationManagerCompat;
+import android.support.v4.app.TaskStackBuilder;
+import android.support.v4.app.NotificationCompat;
+import android.util.Log;
+import android.widget.Toast;
+import java.util.Random;
+
+public class HereYouGoService extends Service
```
cco3: I'd prefer we rename this to `PhysicalWebBroadcastService`

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+        implements PwsClient.ShortenUrlCallbacks,
```
cco3: You already shorten in the activity.  You sholudn't need to shorten here.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+        PwsClient.ResolveScanCallback {
+
+    private static final String TAG = "UrlSharerService";
```
cco3: Pick a more fitting tag

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+    private BluetoothManager mBluetoothManager;
+    private BluetoothLeAdvertiser mBluetoothLeAdvertiser;
+    private static final ParcelUuid URI_BEACON_UUID = ParcelUuid.fromString("0000FEAA-0000-1000-8000-00805F9B34FB");
```
cco3: This seems like it belongs in the advertising utils...but don't move it just yet.  Let's talk about it after you make the other changes.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+        Log.d(TAG, "SERVICE onStartCommand");
+        IntentFilter filter = new IntentFilter(BluetoothAdapter.ACTION_STATE_CHANGED);
+        registerReceiver(mReceiver, filter);
+        // Only call initialize once
+        // We do this because MainActivity launches the service via startService
+        // but it may call it more than once during initialization
+        url = intent.getStringExtra("URL");
+        ourl = intent.getStringExtra("OURL");
```
cco3: Please give these other names like `broadcastUrl` and `displayUrl`

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    // Fires when user swipes away app from the recent apps list
+    public void onTaskRemoved (Intent rootIntent) {
+        Log.d(TAG, "onTaskRemoved");
+        // if (getCanStopSelf()) {
+        //     stopSelf();
+        // }
+        stopSelf();
+        //setCanStopSelf(true);
```
cco3: try to remove all the commented out code

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    @Override
+    public IBinder onBind(Intent intent) {
```
cco3: do we need this?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+        return mLocalBinder;
+    }
+
+    // The callbacks for the ble advertisement events
+    private final AdvertiseCallback mAdvertiseCallback = new AdvertiseCallback() {
```
cco3: Can the outer class just implement AdvertiseCallback?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+    /////////////////////////////////
+    // utilities
+    /////////////////////////////////
+
+    // Kick off the sharing of a URL
+    // We store the long URL
+    // and then ask for its metadata
+    // Once the metadata returns, then we go about either shortening it
+    // and then broadcasting, or just broadcasting if the URL is short enough
+    public void shareUrl(String url) {
```
cco3: remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+        //setLongShareUrl(url);
+        //requestUrlMetaData(url);
+    }
+
+
+    // If the URL is short enough to broadcast
+    private static boolean hasValidUrlLength(String url) {
```
cco3: remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/HereYouGoService.java
```diff
@@ -0,0 +1,332 @@
...
+            Log.d(TAG, "broadcastUrl: " + url);
+            final AdvertiseData advertisementData = getAdvertisementData(url);
+            final AdvertiseSettings advertiseSettings = getAdvertiseSettings();
+            mBluetoothLeAdvertiser.stopAdvertising(mAdvertiseCallback);
+            mBluetoothLeAdvertiser.startAdvertising(advertiseSettings, advertisementData, mAdvertiseCallback);
+        }
+    }
+
+    // Turn on URL broadcasting
+    public void enableUrlBroadcasting() {
```
cco3: do we use this and disableBroadcasting?  If so, what do we use it for?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoMetadata.java
```diff
@@ -0,0 +1,330 @@
...
+                    jsonObj.getInt(TX_POWER_KEY));
+            return bleMetadata;
+        }
+
+        public static BleMetadata fromJsonStr(String jsonStr) throws JSONException {
+            return fromJsonObj(new JSONObject(jsonStr));
+        }
+
+        /*
+        private static RegionResolver getRegionResolver() {
```
cco3: Remove all this unused code

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwoMetadata.java
```diff
@@ -0,0 +1,330 @@
...
+import org.json.JSONObject;
+
+import java.io.ByteArrayOutputStream;
+
+/**
+ * This class holds data about a Physical Web Object and its url.
+ *
+ * A physical web object is any source of a broadcasted url.
+ */
+class PwoMetadata implements Comparable<PwoMetadata> {
```
cco3: Do you actually use this?  If so, you should be using `UrlDevice` instead

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.Collection;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+// Based on code at
+// https://github.com/google/physical-web/blob/master/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PwsClient.java
+public class PwsClient {
```
cco3: Use PwsClient from the collection library instead.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UriBeacon.java
```diff
@@ -0,0 +1,539 @@
...
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.UUID;
+import java.util.Arrays;
+
+/**
+ * Represents a Uri Beacon from Bluetooth LE scan.
+ */
+
+public class UriBeacon {
```
cco3: See org.physical_web.ble.UriBeacon (you shouldn't need this class)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -193,7 +193,7 @@ public void onCreate() {
     initialize();
     restoreCache();
 
-    mNotificationManager.cancelAll();
+    //mNotificationManager.cancelAll();
```
cco3: Just call cancel on the proper ID, don't comment out these lines entirely.

android/PhysicalWeb/app/src/main/AndroidManifest.xml
```diff
@@ -62,6 +75,11 @@
             android:name="org.uribeacon.config.GattService"
             android:exported="false">
         </service>
+	<service
```
cco3: Nit: line this up please (same with the above activity)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/AdvertiseDataUtils.java
```diff
@@ -0,0 +1,161 @@
+/*
+Helper class to simplify Eddystone URL encoding
+*/
+
+package org.physical_web.physicalweb;
```
cco3: Please move this to the ble package

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
+/*
+MainActivity.java
```
cco3: Fix up this comment...add the normal header with the appropriate date

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Handles share intent.
+ */
+public class BroadcastActivity extends Activity {
+
+    private static final String TAG  = "BroadcastActivity";
+    private static final int ENABLE_BLUETOOTH_REQUEST_ID = 1;
+    public static final int MAX_URI_LENGTH = 18;
+    private String displayUrl;
```
cco3: This member shouldn't be needed

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+
+
+    // callbacks
+    /////////////////////////////////
+    @Override
+    protected void onActivityResult (int requestCode, int resultCode, Intent data) {
+      Log.d(TAG, "onActivityResult");
+      if (requestCode == ENABLE_BLUETOOTH_REQUEST_ID && resultCode == -1) {
+        if (checkLE()){
+            preBroadcast();
```
cco3: indents are off here

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+                      UrlShortenerClient.getInstance(this).shortenUrl(displayUrl, urlSetter, TAG);
+                    }
+                } else {
+                  Toast.makeText(this, "Couldn't find URL to broadcast", Toast.LENGTH_LONG).show();
+                }
+            }
+        }
+    }
+
+    private boolean checkLE() {
```
cco3: nit: `BLE` would be better

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+    // callbacks
+    /////////////////////////////////
+    @Override
+    protected void onActivityResult (int requestCode, int resultCode, Intent data) {
+      Log.d(TAG, "onActivityResult");
+      if (requestCode == ENABLE_BLUETOOTH_REQUEST_ID && resultCode == -1) {
+        if (checkLE()){
+            preBroadcast();
+        }
+            finish();
```
cco3: checkLE already calls finish if it returns false (but maybe it shouldn't and that should be left to its caller)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+        Toast.makeText(this, "Turn Bluetooth on to broadcast", Toast.LENGTH_LONG).show();
+        finish();
+      }
+    }
+
+    @Override
+    protected void onStart() {
+        super.onStart();
+        if (Build.VERSION.SDK_INT < 21){
+            Toast.makeText(this, "Bluetooth LE advertising is not supported (API)",
```
cco3: s/(API)/by your operating system/

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+                Toast.LENGTH_LONG).show();
+            return;
+        }
+        if (!checkIfBluetoothIsEnabled()){
+            requestBluetooth();
+        } else {
+            if (checkLE()){
```
cco3: I'm thinking this should be its own method since you have the same code up above for when the permission loop is complete.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+                preBroadcast();
+            }
+                finish();
+        }
+
+        
+    }
+
+    private void preBroadcast(){
```
cco3: Colud you give this another name?  Maybe `parseAndHandleUrl`?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+        Intent intent = getIntent();
+        String action = intent.getAction();
+        String type = intent.getType();
+        if (action.equals(Intent.ACTION_SEND) && (type != null)) {
```
cco3: isn't it always action_send?  isn't it type always text/plain?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+            Log.d(TAG, "content found!");
+            if (type.equals("text/plain")) {
+                Log.d(TAG, "text received");
+                String text = intent.getStringExtra(Intent.EXTRA_TEXT);
+                // Look to see if the text is a URL
+                displayUrl = findUrlInText(text);
+                // If the text is a URL
+                if (displayUrl != null) {
+                    UrlShortenerClient.ShortenUrlCallback urlSetter =
```
cco3: You should create this object in the block where you actually use it.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+        if (bluetoothManager.getAdapter().getBluetoothLeAdvertiser() == null) {
+            Log.d(TAG, "cant advertise");
+            return false;
+        }
+        Log.d(TAG, "can advertise");
+        return true;
+    }
+
+    private void startBroadcastService(String url){
+        Intent inten = new Intent(this, PhysicalWebBroadcastService.class);
```
cco3: s/inten/intent

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+        inten.putExtra("URL", url);
+        inten.putExtra("OURL", displayUrl);
+        startService(inten);
+    }
+
+
+    // Link up to the service so that this activity
```
cco3: Remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        return null;
+    }
+
+    private static boolean hasValidUrlLength(String url) {
+        int uriLength = UriBeacon.uriLength(url);
+        return 0 <= uriLength && uriLength <= MAX_URI_LENGTH;
+    }
+    private static boolean isAsciiUrl(String url) {
```
cco3: space between methods

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,207 @@
...
+        boolean isCompliant = false;
+        try {
+          URI uri = new URI(url);
+          String urlString = uri.toASCIIString();
+          isCompliant = url.equals(urlString);
+        } catch (URISyntaxException e) {
+          // bad url
```
cco3: What do we do here?  toast andfinish "Sorry we do not yet support sharing non-ascii URLs"

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MainActivity.java
```diff
@@ -109,6 +111,7 @@ protected void onResume() {
   protected void onNewIntent(Intent intent) {
     super.onNewIntent(intent);
     setIntent(intent);
+    Log.d(TAG, "new intent BroadcastActivity");
```
cco3: is this right?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    private BluetoothManager mBluetoothManager;
+    private BluetoothLeAdvertiser mBluetoothLeAdvertiser;
+    private static final ParcelUuid URI_BEACON_UUID = ParcelUuid.fromString(
+        "0000FEAA-0000-1000-8000-00805F9B34FB");
+    private boolean amRunning = false;
+    private static final int NOTIFICATION_ID = 6;
+    private NotificationManagerCompat mNotificationManager;
+    private IBinder mLocalBinder;
+    private Handler mHandler = new Handler();
+    private String displayUrl;
```
cco3: Do these need to be members?  If so, they should be prefixed with m

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+
+    private static final String TAG = "PhysicalWebBroadcastService";
+    private BluetoothManager mBluetoothManager;
+    private BluetoothLeAdvertiser mBluetoothLeAdvertiser;
+    private static final ParcelUuid URI_BEACON_UUID = ParcelUuid.fromString(
+        "0000FEAA-0000-1000-8000-00805F9B34FB");
+    private boolean amRunning = false;
+    private static final int NOTIFICATION_ID = 6;
+    private NotificationManagerCompat mNotificationManager;
+    private IBinder mLocalBinder;
```
cco3: Do we need the local binder or the handler?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+ */
+@TargetApi(21)
+public class PhysicalWebBroadcastService extends Service {
+
+    private static final String TAG = "PhysicalWebBroadcastService";
+    private BluetoothManager mBluetoothManager;
+    private BluetoothLeAdvertiser mBluetoothLeAdvertiser;
+    private static final ParcelUuid URI_BEACON_UUID = ParcelUuid.fromString(
+        "0000FEAA-0000-1000-8000-00805F9B34FB");
+    private boolean amRunning = false;
```
cco3: Is this used?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    private IBinder mLocalBinder;
+    private Handler mHandler = new Handler();
+    private String displayUrl;
+    private String shareUrl;
+    private boolean mAmBroadcastingUrl = false;
+    private static final String SHOULD_BROADCAST_URL_KEY = "amBroadcastingUrl";
+    private static final String LONG_SHARE_URL_KEY = "longShareUrl";
+    private static final String URL_TO_BROADCAST_KEY = "urlToBroadcast";
+    private boolean mCanStopSelf = true;
+    public int i1;
```
cco3: Remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    private final BroadcastReceiver mReceiver = new BroadcastReceiver() {
```
cco3: I'd prefer a space before this

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+            Log.d(TAG, "in receiver");
+            String action = intent.getAction();
+
+            if (BluetoothAdapter.ACTION_STATE_CHANGED.equals(action)) {
+                int state = intent.getIntExtra(BluetoothAdapter.EXTRA_STATE, -1);
+                switch (state) {
+                    case BluetoothAdapter.STATE_OFF:
+                        Log.d(TAG, "stop becuase BT off");
+                        stopSelf();
+                        break;
```
cco3: Add a default case

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+                }
+            }
+        }
+    };
+
+    private void initialize() {
```
cco3: I would just put all this in onCreate...it doesn't seem to need its own method

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    @Override
+    public void onDestroy() {
+        Log.d(TAG, "SERVICE onDestroy");
+        unregisterReceiver(mReceiver);
+        disableUrlBroadcasting();
+        super.onDestroy();
+    }
+
+    // Fires when user swipes away app from the recent apps list
+    public void onTaskRemoved (Intent rootIntent) {
```
cco3: Override

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+        Log.d(TAG, "SERVICE onStartCommand");
+        IntentFilter filter = new IntentFilter(BluetoothAdapter.ACTION_STATE_CHANGED);
+        registerReceiver(mReceiver, filter);
+        // Only call initialize once
+        // We do this because MainActivity launches the service via startService
+        // but it may call it more than once during initialization
+        shareUrl = intent.getStringExtra("URL");
+        displayUrl = intent.getStringExtra("OURL");
+        Log.d(TAG, shareUrl);
+        this.broadcastUrl();
```
cco3: `this` shouldn't be needed

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    // Fires when user swipes away app from the recent apps list
+    public void onTaskRemoved (Intent rootIntent) {
+        Log.d(TAG, "onTaskRemoved");
+        stopSelf();
+    }
+
+    @Override
+    public IBinder onBind(Intent intent) {
```
cco3: I don't think we need this

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    // The callbacks for the ble advertisement events
+    private final AdvertiseCallback mAdvertiseCallback = new AdvertiseCallback() {
+
+        // Fires when the URL is successfully being advertised
+        @Override
+        public void onStartSuccess(AdvertiseSettings advertiseSettings) {
+            Log.d(TAG, "URL is broadcasting");
+            mAmBroadcastingUrl = true;
+            createNotification();
+            Toast.makeText(getApplicationContext(), "URL is broadcasting!",
```
cco3: Move this out to the string resource file

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+
+
+
+    /////////////////////////////////
+    // utilities
+    /////////////////////////////////
+
+    // Broadcast via bluetooth the stored URL
+    private void broadcastUrl() {
+        //String urlToBroadcast = getUrlToBroadcast();
```
cco3: Remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+        if (shareUrl != null) {
```
cco3: We should never call broadcastUrl if shareUrl is null

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+            mBluetoothLeAdvertiser.startAdvertising(advertiseSettings,
+                advertisementData, mAdvertiseCallback);
+        }
+    }
+
+    // Turn off URL broadcasting
+    public void disableUrlBroadcasting() {
+        Log.d(TAG, "disableUrlBroadcasting");
+        mBluetoothLeAdvertiser.stopAdvertising(mAdvertiseCallback);
+        mNotificationManager.cancel(10);
```
cco3: Use a constant

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final ParcelUuid URI_BEACON_UUID = ParcelUuid.fromString(
+        "0000FEAA-0000-1000-8000-00805F9B34FB");
+    private boolean amRunning = false;
+    private static final int NOTIFICATION_ID = 6;
+    private NotificationManagerCompat mNotificationManager;
+    private IBinder mLocalBinder;
+    private Handler mHandler = new Handler();
+    private String displayUrl;
+    private String shareUrl;
+    private boolean mAmBroadcastingUrl = false;
```
cco3: This doesn't seem to be used for anything.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final String SHOULD_BROADCAST_URL_KEY = "amBroadcastingUrl";
```
cco3: This doesn't seem to be used

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final String LONG_SHARE_URL_KEY = "longShareUrl";
```
cco3: same here

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final String URL_TO_BROADCAST_KEY = "urlToBroadcast";
```
cco3: same

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+        Log.d(TAG, "SERVICE onStartCommand");
+        IntentFilter filter = new IntentFilter(BluetoothAdapter.ACTION_STATE_CHANGED);
+        registerReceiver(mReceiver, filter);
+        // Only call initialize once
+        // We do this because MainActivity launches the service via startService
+        // but it may call it more than once during initialization
+        shareUrl = intent.getStringExtra("URL");
```
cco3: Use constants

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    // Turn off URL broadcasting
+    public void disableUrlBroadcasting() {
+        Log.d(TAG, "disableUrlBroadcasting");
+        mBluetoothLeAdvertiser.stopAdvertising(mAdvertiseCallback);
+        mNotificationManager.cancel(10);
+        mAmBroadcastingUrl = false;
+    }
+
+    // Generate the advertising bytes for the given URL
+    private AdvertiseData getAdvertisementData(String uri) {
```
cco3: Could this go into the advertiseutils class?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+            PendingIntent.FLAG_UPDATE_CURRENT);
+        registerReceiver(stopServiceReceiver, new IntentFilter("myFilter"));
+        PendingIntent pIntent = PendingIntent.getBroadcast(this, 0, new Intent("myFilter"),
+            PendingIntent.FLAG_UPDATE_CURRENT);
+        NotificationCompat.Builder mBuilder = new NotificationCompat.Builder(this)
+                .setSmallIcon(R.drawable.ic_leak_add_white_24dp)
+                .setContentTitle("Physical Web is broadcasting a URL")
+                .setContentText(displayUrl)
+                .setOngoing(true)
+                .addAction(2, "Stop", pIntent);
```
cco3: Make this a string resource.  Also, don't hardcode the resource id

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+        mBuilder.setContentIntent(resultPendingIntent);
+
+        NotificationManager mNotificationManager = (NotificationManager) getSystemService(
+            Context.NOTIFICATION_SERVICE);
+        mNotificationManager.notify(10, mBuilder.build());
```
cco3: use a constant here

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    protected BroadcastReceiver stopServiceReceiver = new BroadcastReceiver() {
```
cco3: If this needs to be declared outside the method where it's used, move this up to the top and make it a member.  Use the m prefix.

However, I'm thinking it'd make the most sense just to create it in `createNotification`.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -313,7 +313,7 @@ private void updateNotifications() {
     // If no beacons have been found
     if (pwPairs.size() == 0) {
       // Remove all existing notifications
-      mNotificationManager.cancelAll();
+      //mNotificationManager.cancelAll();
```
cco3: This line is important, just cancel on the specific ID

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+BroadcastActivity.java
```
cco3: this comment should be moved down just above the beginning of the class

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Handles share intent.
+ */
+public class BroadcastActivity extends Activity {
+
+    private static final String TAG  = "BroadcastActivity";
+    private static final int ENABLE_BLUETOOTH_REQUEST_ID = 1;
+    public static final int MAX_URI_LENGTH = 18;
+    private static final String SHARE_CONST = "URL";
```
cco3: Remove these, make the ones in the service public and use them.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,205 @@
...
+ * Handles url broadcasting.
+ */
+@TargetApi(21)
+public class PhysicalWebBroadcastService extends Service {
+
+    private static final String TAG = "PhysicalWebBroadcastService";
+    private BluetoothManager mBluetoothManager;
+    private BluetoothLeAdvertiser mBluetoothLeAdvertiser;
+    private static final int NOTIFICATION_ID = 6;
+    private static final String SHARE_CONST = "URL";
```
cco3: Rename these to `BROADCAST_URL_KEY` and `DISPLAY_URL_KEY`

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,217 @@
...
+    private static final String DISP_CONST = "OURL";
+    private String url;
+
+
+    // callbacks
```
cco3: You lost the first half of this formatted comment...I'm fine with you just removing the whole thing.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,217 @@
...
+    /////////////////////////////////
+    @Override
+    protected void onActivityResult (int requestCode, int resultCode, Intent data) {
+        Log.d(TAG, "onActivityResult");
+        if (requestCode == ENABLE_BLUETOOTH_REQUEST_ID && resultCode == -1) {
+            checkBleAndStart();
+            finish();
```
cco3: You can move finish outside the if statement

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,217 @@
...
+                UrlShortenerClient.ShortenUrlCallback urlSetter =
+                new UrlShortenerClient.ShortenUrlCallback() {
+                    @Override
+                    public void onUrlShortened(String newUrl) {
+                        Log.d(TAG, "shortening success");
+                        startBroadcastService(url, newUrl);
+                    }
+                    @Override
+                    public void onError(String oldUrl) {
+                        Log.d(TAG, "shortening error");
```
cco3: You should make a toast and finish

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,205 @@
+/*
+HereYouGoService.java
```
cco3: Put the normal copyright header here.  Move the description down just above the class definition.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,205 @@
...
+
+            if (BluetoothAdapter.ACTION_STATE_CHANGED.equals(action)) {
+                int state = intent.getIntExtra(BluetoothAdapter.EXTRA_STATE, -1);
+                switch (state) {
+                    case BluetoothAdapter.STATE_OFF:
+                        Log.d(TAG, "stop becuase BT off");
+                        stopSelf();
+                        break;
+                    default:
+                        break;
```
cco3: Don't break in a default case.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,205 @@
...
+            PendingIntent.FLAG_UPDATE_CURRENT);
+        registerReceiver(stopServiceReceiver, new IntentFilter("myFilter"));
+        PendingIntent pIntent = PendingIntent.getBroadcast(this, 0, new Intent("myFilter"),
+            PendingIntent.FLAG_UPDATE_CURRENT);
+        NotificationCompat.Builder mBuilder = new NotificationCompat.Builder(this)
+                .setSmallIcon(R.drawable.ic_leak_add_white_24dp)
+                .setContentTitle(getString(R.string.broadcasting2))
+                .setContentText(mdisplayUrl)
+                .setOngoing(true)
+                .addAction(2, "Stop", pIntent);
```
cco3: Use a string resource for "Stop" .  Also use R.id.<identifier> to get the id here.

android/PhysicalWeb/app/src/main/res/values/strings.xml
```diff
@@ -21,11 +21,17 @@
...
     <string name="notification_view_more_beacons">VIEW MORE</string>
     <string name="config_url_read_error">Error Reading URL</string>
-
+    <string name="bt_on">Turn Bluetooth on to broadcast</string>
+    <string name="ble_os_error">Bluetooth LE advertising is not supported by your OS</string>
+    <string name="no_url_error">Couldn\'t find URL to broadcast</string>
+    <string name="ble_error">Bluetooth LE advertising is not supported on this device</string>
+    <string name="nonascii_error">Sorry we don\'t support non-ascii URLs</string>
+    <string name="broadcasting">URL is broadcasting!</string>
+    <string name="broadcasting2"> Physical Web is broadcasting a URL</string>
```
cco3: Please give these two more distinctive names

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/ble/AdvertiseDataUtils.java
```diff
@@ -0,0 +1,203 @@
+/*
+Helper class to simplify Eddystone URL encoding
```
cco3: Add the normal copyright header.  Move this comment just above the class definition.

cco3:
Looking good, just a few things to cleanup.

nishadg246:
pushed again

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,216 @@
...
+import android.os.IBinder;
+import android.support.v4.app.NotificationCompat;
+import android.support.v4.app.NotificationManagerCompat;
+import android.support.v4.app.TaskStackBuilder;
+import android.util.Log;
+import android.widget.Toast;
+import org.physical_web.physicalweb.ble.AdvertiseDataUtils;
+
+/**
+    * Shares URLs via bluetooth.
```
cco3: 2 spaces

cco3:
Just one indent off, and broadcast/broadcast2 still needs a better name.  @nondebug, I'm pretty much done if you want to take a look.

cco3:
LGTM, thanks!

android/PhysicalWeb/app/src/main/AndroidManifest.xml
```diff
@@ -33,6 +33,19 @@
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
         </activity>
+        <activity
+            android:name=".BroadcastActivity"
+            android:label="@string/app_name"
+            android:launchMode="singleTask"
+            android:theme="@android:style/Theme.NoDisplay"
+            android:excludeFromRecents="true"
+            android:taskAffinity="app.broadcast">
```
nondebug: With launchMode="singleTask" there's no reason to set taskAffinity (the activity will always be created in its own task)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+        // Look to see if the text is a URL
+        url = findUrlInText(text);
+        // If the text is a URL
+        if (url != null) {
+            if (hasValidUrlLength(url) && isAsciiUrl(url)) {
+                // Set the url if we can
+                Log.d(TAG, "valid length");
+                startBroadcastService(url, url);
+            } else {
+                Log.d(TAG, "needs shortening");
```
nondebug: Will we get here if the URL is non-ASCII?

cco3: Yes, and those URLs still require "shortening" so that it will fit into the eddystone-url format

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+        if (urls.size() > 0) {
+            return urls.get(0);
+        }
+
+        return null;
+    }
+
+    private static boolean hasValidUrlLength(String url) {
+        int uriLength = UriBeacon.uriLength(url);
+        return 0 <= uriLength && uriLength <= MAX_URI_LENGTH;
```
nondebug: Is zero a valid URL length?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+        if (urls.size() > 0) {
+            return urls.get(0);
+        }
+
+        return null;
+    }
+
+    private static boolean hasValidUrlLength(String url) {
+        int uriLength = UriBeacon.uriLength(url);
```
nondebug: Let's try to remove the dependency on UriBeacon.  The implementation of uriLength is here:

https://github.com/google/uribeacon/blob/uribeacon-final/android-uribeacon/uribeacon-library/src/main/java/org/uribeacon/beacon/UriBeacon.java#L227

Please add it to AdvertiseDataUtils.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * This is the main entry point for the broadcasting service.
+ * Handles the share intent.
+ */
+public class BroadcastActivity extends Activity {
+
+    private static final String TAG  = "BroadcastActivity";
+    private static final int ENABLE_BLUETOOTH_REQUEST_ID = 1;
+    public static final int MAX_URI_LENGTH = 18;
+    private String url;
```
nondebug: Does this still need to be a member?  It appears to be unused except in parseAndHandleUrl where the value is set.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+        } else {
+            Toast.makeText(this, getString(R.string.bt_on), Toast.LENGTH_LONG).show();
+        }
+        finish();
+    }
+
+    @Override
+    protected void onStart() {
+        super.onStart();
+        if (Build.VERSION.SDK_INT < 21){
```
nondebug: Please add a comment indicating why this check is necessary

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+            finish();
+        }
+    }
+
+    private void parseAndHandleUrl(){
+        Intent intent = getIntent();
+        String action = intent.getAction();
+        String type = intent.getType();
+        String text = intent.getStringExtra(Intent.EXTRA_TEXT);
+        // Look to see if the text is a URL
```
nondebug: nit: this actually checks if the text contains at least one URL

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    private void checkBleAndStart(){
+        if (checkAndWarnBleAdvertiseCapability()) {
+            parseAndHandleUrl();
+        }
+    }
+    private boolean checkAndWarnBleAdvertiseCapability() {
+        if (!hasBleAdvertiseCapability()) {
+            Toast.makeText(this, getString(R.string.ble_error),
```
nondebug: You could fold this into checkBleAndStart:

if (hasBleAdvertiseCapability())
  parseAndHandleUrl();
else
  Toast.makeText(...).show();

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    private boolean isAsciiUrl(String url) {
+        boolean isCompliant = false;
+        try {
+          URI uri = new URI(url);
+          String urlString = uri.toASCIIString();
+          isCompliant = url.equals(urlString);
+        } catch (URISyntaxException e) {
+            Toast.makeText(this, getString(R.string.nonascii_error), Toast.LENGTH_LONG).show();
```
nondebug: "isAsciiUrl" suggests this function does nothing more than check the URL and return a boolean, but it can also create a toast as a side effect.  Could you either move the toast logic to the caller or rename the function to make this behavior more apparent?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,216 @@
...
+  * Shares URLs via bluetooth.
+  * Also interfaces with PWS to shorten URLs that are too long for Eddystone URLs.
+  * Lastly, it surfaces a persistent notification whenever a URL is currently being broadcast.
+**/
+
+@TargetApi(21)
+public class PhysicalWebBroadcastService extends Service {
+
+    private static final String TAG = "PhysicalWebBroadcastService";
+    private BluetoothManager mBluetoothManager;
```
nondebug: mBluetoothManager is only used to get the BluetoothLeAdvertiser, so it doesn't need to be a member here

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,216 @@
...
+    private BluetoothLeAdvertiser mBluetoothLeAdvertiser;
+    private static final int NOTIFICATION_ID = 6;
```
nondebug: Since we have multiple notifications in this app, could you rename this to make it clear this is the broadcast notification?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,216 @@
...
+    public static final String BROADCAST_URL_KEY = "URL";
+    public static final String DISPLAY_URL_KEY = "OURL";
```
nondebug: Could you make these keys more descriptive?  Why not "broadcastUrl" and "displayUrl"?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,216 @@
...
+    private NotificationManagerCompat mNotificationManager;
+    private IBinder mLocalBinder;
+    private Handler mHandler = new Handler();
+    private String mdisplayUrl;
+    private String shareUrl;
```
nondebug: Please rename
mdisplayUrl -> mDisplayUrl
shareUrl -> mShareUrl

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,216 @@
...
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            Log.d(TAG, "in receiver");
+            String action = intent.getAction();
+
+            if (BluetoothAdapter.ACTION_STATE_CHANGED.equals(action)) {
+                int state = intent.getIntExtra(BluetoothAdapter.EXTRA_STATE, -1);
+                switch (state) {
+                    case BluetoothAdapter.STATE_OFF:
+                        Log.d(TAG, "stop becuase BT off");
```
nondebug: fix spelling becuase -> because

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,216 @@
...
+    // Fires when user swipes away app from the recent apps list
+    @Override
+    public void onTaskRemoved (Intent rootIntent) {
+        Log.d(TAG, "onTaskRemoved");
+        stopSelf();
+    }
+
+    @Override
+    public IBinder onBind(Intent intent) {
+        return mLocalBinder;
```
nondebug: mLocalBinder is never set, so it will always be null.  Might as well remove the member and return null instead.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -221,6 +221,11 @@ public boolean onUnbind(Intent intent) {
   public void onRebind(Intent intent) {
     mIsBound = true;
   }
+  private void cancelNotifs() {
```
nondebug: Please rename to cancelNotifications

android/PhysicalWeb/app/src/main/res/values/strings.xml
```diff
@@ -21,16 +20,23 @@
...
+    <string name="url_getting_started">https://google.github.io/physical-web/mobile/android/getting-started.html</string>
     <string name="config_no_beacons_found">No beacons found</string>
     <string name="notification_view_more_beacons">VIEW MORE</string>
     <string name="config_url_read_error">Error Reading URL</string>
-
+    <string name="bt_on">Turn Bluetooth on to broadcast</string>
+    <string name="ble_os_error">Bluetooth LE advertising is not supported by your OS</string>
+    <string name="no_url_error">Couldn\'t find URL to broadcast</string>
+    <string name="ble_error">Bluetooth LE advertising is not supported on this device</string>
+    <string name="nonascii_error">Sorry we don\'t support non-ascii URLs</string>
```
nondebug: Could you reword this to avoid using "we"?  Also, ASCII should be capitalized.

android/PhysicalWeb/app/src/main/res/values/strings.xml
```diff
@@ -21,16 +20,23 @@
...
+    <string name="url_broadcast">URL is broadcasting!</string>
+    <string name="broadcast_notif"> Physical Web is broadcasting a URL</string>
```
nondebug: The leading space will get ignored and won't show up in the notification, but could you remove it anyway?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
nondebug: What happens if we get two toasts at once?  (isAsciiUrl toast and the shorten_error toast below)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BroadcastActivity.java
```diff
@@ -0,0 +1,205 @@
...
+        int uriLength = UriBeacon.uriLength(url);
+        return 0 < uriLength && uriLength <= MAX_URI_LENGTH;
+    }
+
+    private boolean isAsciiUrl(String url) {
+        boolean isCompliant = false;
+        try {
+          URI uri = new URI(url);
+          String urlString = uri.toASCIIString();
+          isCompliant = url.equals(urlString);
```
nondebug: Should we emit a toast when isCompliant == false?

cco3: I don't think so...we will just throw it into a shortener.

nishadg246:
more changes

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    private boolean checkAndHandleAsciiUrl(String url) {
+        boolean isCompliant = false;
+        try {
+          URI uri = new URI(url);
+          String urlString = uri.toASCIIString();
+          isCompliant = url.equals(urlString);
+        } catch (URISyntaxException e) {
+            Toast.makeText(this, getString(R.string.nonascii_error), Toast.LENGTH_LONG).show();
```
nondebug: Is this toast necessary?  It says "non-ASCII URLs not supported" but in fact we do support them.

cco3: I think I might have asked him to add that.  whoops!

nishadg246: How should I handle the exception

nondebug: I'm not sure.  What inputs would cause you to trigger that exception?  If it's only for non-ASCII URLs then it's probably fine to catch and ignore it.  But if it's something that would also trip up the shortener then we should show a toast and not try to shorten it.

cco3: Oh, this is for things that aren't URLs....depending on how airtight the search for a URL is, you could just throw a runtime error here.

nishadg246:
fixed

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,253 @@
...
+            PendingIntent.FLAG_UPDATE_CURRENT);
+        NotificationCompat.Builder mBuilder = new NotificationCompat.Builder(this)
+                .setSmallIcon(R.drawable.ic_leak_add_white_24dp)
+                .setContentTitle(getString(R.string.broadcast_notif))
+                .setContentText(mDisplayUrl)
+                .setOngoing(true)
+                .addAction(android.R.drawable.ic_menu_close_clear_cancel,
+                    getString(R.string.stop), pIntent);
+        mBuilder.setContentIntent(resultPendingIntent)
+;
```
nondebug: this semicolon should be on the previous line

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebBroadcastService.java
```diff
@@ -0,0 +1,253 @@
...
+
+    /////////////////////////////////
+    // callbacks
+    /////////////////////////////////
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        Log.d(TAG, "SERVICE onCreate");
+        BluetoothManager mBluetoothManager = (BluetoothManager) getApplicationContext()
```
nondebug: Please rename to bluetoothManager (the m suggests it's a private member)

nondebug:
Please address the remaining comments, then it LGTM

nishadg246:
done

PR #683, opened 2016-06-07
Warn users when URL is filtered out

hayesjordan:


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BeaconConfigFragment.java
```diff
@@ -444,6 +448,36 @@ public void onError(String oldUrl) {
...
+          });
+        }
+
+        @Override
+        public void onPwsResult(PwsResult pwsResult) {
+        }
+
+        @Override
+        public void onPwsResultError(Collection<String> urls, int httpResponseCode, Exception e) {
+          if (urls.iterator().next().contains("https")){
```
cco3: startsWith

nondebug: We should toLower the URL before checking the scheme so we can match (eg) HTTPS://www.google.com

android/PhysicalWeb/app/src/main/res/values/strings.xml
```diff
@@ -64,5 +64,7 @@
     <string name="url_charset_error">Non-ASCII characters unsupported in URL. Please use a shortener.</string>
     <string name="error_bluetooth_support">Device does not support Bluetooth</string>
     <string name="error_api_key_no_longer_available">Google API key no longer available, please update settings.</string>
+    <string name="pws_filtered_url_unknown">Other clients may filter your URL.</string>
+    <string name="pws_filtered_url_http">Other clients may filter your URL because it is not https.</string>
```
nondebug: Let's capitalize HTTPS in the warning string

nondebug:
lgtm

PR #687, opened 2016-06-10
Bug fix: listview not displaying

hayesjordan:


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -287,6 +289,7 @@ private void startScanningDisplay(long scanStartTime, boolean hasResults) {
     long elapsedMillis = new Date().getTime() - scanStartTime;
     if (elapsedMillis < FIRST_SCAN_TIME_MILLIS
         || (elapsedMillis < SECOND_SCAN_TIME_MILLIS && !hasResults)) {
+      mNearbyDeviceAdapter.clear();
```
nondebug: I think this only works because clearing the ListAdapter triggers the empty view in the containing ListFragment.  It's suspicious to me that we ever explicitly change the visibility of the ListView since this is supposed to be handled automatically, see https://developer.android.com/reference/android/app/ListFragment.html

If it's not possible to rely on this behavior then it's fine to do it manually, but we should change the id of the empty view in fragment_nearby_beacons.xml so it's not "@id/android:empty" since according to the docs that's what's causing the unexpected visibility changes.

hayesjordan:
Updated per discussion.

android/PhysicalWeb/app/src/main/res/layout/fragment_nearby_beacons.xml
```diff
@@ -9,8 +9,7 @@
                     android:background="#ffffffff">
 
         <ListView
-            android:id="@android:id/list"
-            android:alpha="0"
+            android:id="@id/android:list"
```
nondebug: This was actually correct before; the id for the empty view should be changed to "@android:id/empty"

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -483,6 +483,7 @@ public void restartScan() {
       urlDeviceDiscoverer.stopScan();
     }
     mScanStartTime = new Date().getTime();
+    mPwCollection = new PhysicalWebCollection();
```
nondebug: Why is this necessary?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -106,7 +106,6 @@ public void run() {
     public void run() {
       Log.d(TAG, "running second scan timeout");
       emptyGroupIdQueue();
-      showListView();
```
nondebug: This previously would call mSwipeRefreshWidget.setRefreshing(false), is that no longer necessary?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
hayesjordan: This clears the cached data. Otherwise when you manually refresh you still get the cached data along with scanning for new beacons.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
hayesjordan: Correct. It only needs to be set to false once after a manual refresh which occurs in the first timeout.

nondebug:
lgtm

PR #688, opened 2016-06-13
Preventing Repeated Resolves

nishadg246:
When a URL is not resolved by the PWS, the client
will continue to attempt to fetch a result unnecessarily


nishadg246:
@nondebug

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -368,6 +370,7 @@ public void fetchPwsResults(final PwsResultCallback pwsResultCallback,
       @Override
       public void onPwsResult(PwsResult pwsResult) {
         addMetadata(pwsResult);
+        System.out.println("resolve: result "+ pwsResult.getSiteUrl());
```
nondebug: Please remove the logging, otherwise lgtm

PR #689, opened 2016-06-13
Preventing Repeated Fetches and Resolves

nishadg246:
When a URL is not resolved by the PWS, the client
will continue to attempt to fetch a result unnecessarily.
Also, we refetch whenever we see a new beacon rather than
only when we see a beacon not seen during the current scan

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
@@ -257,7 +257,11 @@ public class UrlDeviceDiscoveryService extends Service
 
   @Override
   public void onUrlDeviceDiscovered(UrlDevice urlDevice) {
-    mPwCollection.addUrlDevice(urlDevice);
+    // check if device has already been discovered
+    if (mPwCollection.addUrlDevice(urlDevice)) {
+      return;
+    }
+    // if we have found a new device
     mPwCollection.fetchPwsResults(new PwsResultCallback() {
       long mPwsTripTimeMillis = 0;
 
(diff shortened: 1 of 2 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -257,47 +257,49 @@ public void onDestroy() {
...
-      }
-
-      @Override
-      public void onError(int httpResponseCode, Exception e) {
-        Log.d(TAG, "PwsResultError: " + httpResponseCode + " ", e);
-        triggerCallback();
-      }
-    });
+    boolean alreadyFound = mPwCollection.addUrlDevice(urlDevice);
+    if (!alreadyFound) {
```
nondebug: How about:

if (mPwCollection.addUrlDevice(urlDevice)) {
  // Device is already present in collection.
  return;
}
mPwCollection.fetchPwsResults(...

nondebug: I missed that we still want to triggerCallback() below

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -52,14 +53,17 @@ public PhysicalWebCollection() {
     mIconUrlToIconMap = new HashMap<>();
     mPendingBroadcastUrls = new HashSet<>();
     mPendingIconUrls = new HashSet<>();
+    mFailedResolveUrls = new HashSet<>();
   }
 
   /**
    * Add a UrlDevice to the collection.
    * @param urlDevice The UrlDevice to add.
```
nondebug: Please extend the javadoc to document the new @return value

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -52,14 +53,18 @@ public PhysicalWebCollection() {
...
+   * @return alreadyFound Had the device already existed in the map
```
nondebug: @return doesn't need a variable name as part of the description, so I'd reword this as "@return true if the device already existed in the map" (or similar)

nondebug:
lgtm

PR #690, opened 2016-06-14
Add custom PWS endpoint to settings page

hayesjordan:


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -46,37 +53,100 @@ public void onCreate(Bundle savedInstanceState) {
...
+    if (homeBtn != null) {
+      OnClickListener dismissDialogClickListener = new OnClickListener() {
+          @Override
+          public void onClick(View v) {
+            updatePwsPreference();
+            updatePwsList();
+            dialog.dismiss();
+          }
+      };
+      ((ViewGroup) homeBtn.getParent().getParent()).setOnClickListener(dismissDialogClickListener);
```
nondebug: Can you use dialog.findViewById() to get this View rather than walking the tree?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -107,6 +107,7 @@ public void run() {
       emptyGroupIdQueue();
       mSecondScanComplete = true;
       mSwipeRefreshWidget.setRefreshing(false);
+      mScanningAnimationTextView.setAlpha(0f);
```
nondebug: Would it make sense to call this everywhere we already call mSwipeRefreshWidget.setRefreshing(false)?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -46,37 +53,100 @@ public void onCreate(Bundle savedInstanceState) {
     getActivity().getActionBar().setDisplayHomeAsUpEnabled(true);
     // Load the preferences from an XML resource
     addPreferencesFromResource(R.xml.settings);
+    updatePwsList();
+  }
+
+  @Override
+  public void onResume() {
+    super.onResume();
+    getActivity().getActionBar().setTitle(R.string.title_settings);
```
nondebug: Would it work to do this in onCreate?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -91,48 +89,103 @@ private static int getCurrentPwsEndpointVersion(Context context){
...
+  }
+
+  private static String readCustomPwsEndpointApiKey(Context context){
+    SharedPreferences sharedPref = PreferenceManager.getDefaultSharedPreferences(context);
+    return sharedPref.getString(context.getString(R.string.custom_pws_api_key_key), "");
+  }
+
+  public static String convertToSettingsEndpointFormat(String pwsUrl, int pwsVerion,
+                                                       String apiKey){
+    return pwsUrl + "," + pwsVerion + "," + apiKey;
```
nondebug: Is it possible to set a custom PWS URL that contains a comma character?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -58,6 +58,8 @@
   private static final String PWSTRIPTIME_KEY = "pwstriptime";
   private static final RegionResolver REGION_RESOLVER = new RegionResolver();
 
+  private static boolean userHasBeenWarned = false;
```
nondebug: Is there somewhere else we could store this bit?  We should try not to store application state in the Utils class.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
hayesjordan: I am not sure what the Id of this element is or exactly what the grandparent of the home button would be.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
hayesjordan: I think it only needs to be called when the animation has stopped and there are no results. If we call it in the first timeout there would be results so its not needed. Also in `startScanningDisplay ` I don't think its need because if it hits the else the timeouts will be firing soon. I can definitely add it to the `startScanningDisplay ` if you think its makes it clearer.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -91,48 +89,103 @@ private static int getCurrentPwsEndpointVersion(Context context){
...
+  }
+
+  public static String convertToSettingsEndpointFormat(String pwsUrl, int pwsVerion,
+                                                       String apiKey){
+    return pwsUrl + "," + pwsVerion + "," + apiKey;
   }
 
   public static void setPwsEndpoint(Context context, PwsClient pwsClient){
-    String apiKey = getApiKey(context);
+    String apiKey = getCurrentPwsEndpointApiKey(context);
```
nondebug: Could you factor out the body of these setPwsEndpoint methods to avoid the duplicated code?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
hayesjordan: For consistency I think it should be done here at least for now. All the other fragments set the title in the `onResume` method.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -91,48 +89,103 @@ private static int getCurrentPwsEndpointVersion(Context context){
...
     String pwsEndpoint = getCurrentPwsEndpointUrl(context);
     int apiVersion = getCurrentPwsEndpointVersion(context);
-    // Check for case where last saved settings was the Google PWS, but user has reinstalled
-    // the app without the API key.
+    if (apiVersion >= 2 && apiKey.isEmpty() && !userHasBeenWarned){
```
nondebug: This block could be part of the block below it, eg:
```
if (apiVersion >= 2 && apiKey.isEmpty()) {
  if (!userHasBeenWarned) {
    // warn user
  }
  pwsEndpoint = ...
}
```

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
hayesjordan: I'm not sure where the best place to put this would be. As of right now the Utils class is the only area that knows about this `Toast`.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -46,37 +53,100 @@ public void onCreate(Bundle savedInstanceState) {
...
+  @Override
+  public void onPrepareOptionsMenu(Menu menu) {
+    super.onPrepareOptionsMenu(menu);
+    menu.findItem(R.id.action_config).setVisible(false);
+    menu.findItem(R.id.action_about).setVisible(false);
+    menu.findItem(R.id.action_settings).setVisible(false);
+  }
+
+  @Override
+  public boolean onPreferenceTreeClick(PreferenceScreen preferenceScreen, Preference preference){
```
nondebug: Insert one space before the opening curly brace (please also fix the other occurrences)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -46,37 +53,100 @@ public void onCreate(Bundle savedInstanceState) {
...
+    super.onPreferenceTreeClick(preferenceScreen, preference);
+
+    if (preference instanceof PreferenceScreen){
+      setupHomeButton(((PreferenceScreen) preference));
```
nondebug: Please remove the unnecessary parens

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -46,37 +53,100 @@ public void onCreate(Bundle savedInstanceState) {
...
+  }
+
+  private void setupHomeButton(PreferenceScreen preference){
+    final Dialog dialog = preference.getDialog();
+    dialog.getActionBar().setDisplayHomeAsUpEnabled(true);
+
+    View homeBtn = dialog.findViewById(android.R.id.home);
+    if (homeBtn != null) {
+      OnClickListener dismissDialogClickListener = new OnClickListener() {
+          @Override
```
nondebug: Indent one level (2 spaces) for the function body

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -46,37 +53,100 @@ public void onCreate(Bundle savedInstanceState) {
...
-
-  @Override
-  public void onPrepareOptionsMenu(Menu menu) {
-    super.onPrepareOptionsMenu(menu);
-    menu.findItem(R.id.action_config).setVisible(false);
-    menu.findItem(R.id.action_about).setVisible(false);
-    menu.findItem(R.id.action_settings).setVisible(false);
+  private void updatePwsPreference(){
+    ListPreference listPreference = (ListPreference) findPreference(
+      getString(R.string.pws_endpoint_setting_key));
```
nondebug: Indent two levels (4 spaces) for a line continuation

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -67,15 +69,11 @@ private static int getGoogleApiKeyResourceId(Context context){
...
-  private static String getApiKey(Context context){
-    int resourceId = getGoogleApiKeyResourceId(context);
-    return resourceId != 0 ? context.getString(resourceId) : "";
-  }
-
   private static String getCurrentPwsEndpoint(Context context){
     String defaultEndpoint = isGoogleApiKeyAvailable(context) ?
-      convertToSettingsEndpointFormat(GOOGLE_ENDPOINT, GOOGLE_ENDPOINT_VERSION) :
-      convertToSettingsEndpointFormat(PROD_ENDPOINT, PROD_ENDPOINT_VERSION);
+      convertToSettingsEndpointFormat(GOOGLE_ENDPOINT, GOOGLE_ENDPOINT_VERSION,
```
nondebug: This line is pretty verbose, could you switch to an if/else instead for clarity?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
hayesjordan: It is possible. What would be a better separator?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
nondebug: What view are you trying to set the OnClickListener for?  It's not clear from the code.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
nondebug: You could use the NUL character (or any other non-printable char) since it should never appear in a valid URL.  Java strings aren't NUL-terminated.

Also I just noticed pwsVerion is missing an S.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
nondebug: I see what you mean, it only has any visible effect in this case since in all other cases the empty view is hidden.  It's fine with me to leave it as-is since you'll be replacing it soon with a better "no results" empty view.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
nondebug: SGTM

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -91,48 +89,103 @@ private static int getCurrentPwsEndpointVersion(Context context){
...
+
+  public static String getCustomPwsEndpoint(Context context){
+    return convertToSettingsEndpointFormat(readCustomPwsEndpointUrl(context),
+                                           readCustomPwsEndpointVersion(context),
+                                           readCustomPwsEndpointApiKey(context));
+  }
+
+  public static void deleteSharedPreferenceKey(Context context, String key){
+    SharedPreferences sharedPref = PreferenceManager.getDefaultSharedPreferences(context);
+    SharedPreferences.Editor editor = sharedPref.edit();
```
nondebug: You can make Conley happy by chaining these calls:
```
SharedPreferences.Editor editor = sharedPref.edit()
  .remove(key)
  .apply();
```

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
nondebug: If setPwsEndpoint returned an error code, the caller could check it and call Utils.warnUserOnMissingApiKey() or something.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
nondebug: *at least 4 spaces

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
hayesjordan: It is the view that holds the back button that appears next to the title. There is a weird bug in Android where nested PreferenceScreens don't show the back button and `dialog.getActionBar().setDisplayHomeAsUpEnabled(true);` doesn't add the functionality to go back like it does normally with fragments.

nondebug: This is pretty brittle as it relies on details of the ActionBar layout that we don't have control over.  What happens if we don't setDisplayHomeAsUpEnabled(true)?  Can you still navigate back using the system Back button?

hayesjordan: Yes the system back button still works if you do `setDisplayHomeAsUpEnabled(true)`. I have also added the functionality that the back button updates the Physical Web Service list. I'm not sure if there is a better way to do this.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
nondebug: Apparently onPreferenceTreeClick is deprecated and we should use onPreferenceStartFragment instead.

https://developer.android.com/reference/android/preference/PreferenceActivity.html#onPreferenceStartFragment(android.preference.PreferenceFragment, android.preference.Preference)

hayesjordan:
Updated.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -16,10 +16,20 @@
 
 package org.physical_web.physicalweb;
 
+import android.content.SharedPreferences;
```
nondebug: Could you remove the unused imports?

hayesjordan: Yes, sorry forgot to run the check. Doing so now with let you know when it is good.

hayesjordan: Removed and the check passed.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/SettingsFragment.java
```diff
@@ -79,4 +69,76 @@ public void onPrepareOptionsMenu(Menu menu) {
...
+      Utils.setPwsEndpointPreference(getActivity(),
+                                     Utils.getDefaultPwsEndpointPreferenceString(getActivity()));
+      Toast.makeText(getActivity(), R.string.error_pws_endpoint_not_configured_properly,
+                     Toast.LENGTH_SHORT).show();
+    }
+
+
+  }
+
+  public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
```
nondebug: add @Override annotation

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -57,82 +57,170 @@
   private static final String TXPOWER_KEY = "tx";
   private static final String PWSTRIPTIME_KEY = "pwstriptime";
   private static final RegionResolver REGION_RESOLVER = new RegionResolver();
+  private static final String SEPARATOR = "\0";
+  private static boolean userHasBeenWarned = false;
+
+  public static class PwsEndpoint {
```
nondebug: Could this be private?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -57,82 +57,170 @@
...
+  }
+
+  private static String readCustomPwsEndpointApiKey(Context context) {
+    SharedPreferences sharedPref = PreferenceManager.getDefaultSharedPreferences(context);
+    return sharedPref.getString(context.getString(R.string.custom_pws_api_key_key), "");
+  }
+
+
+
+  public static String convertToSettingsEndpointFormat(String pwsUrl, int pwsVersion,
```
nondebug: Please add javadocs for the public methods

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -57,82 +57,170 @@
   private static final String TXPOWER_KEY = "tx";
   private static final String PWSTRIPTIME_KEY = "pwstriptime";
   private static final RegionResolver REGION_RESOLVER = new RegionResolver();
+  private static final String SEPARATOR = "\0";
+  private static boolean userHasBeenWarned = false;
```
nondebug: unused, please remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
nondebug: (@override methods don't usually need javadoc)

nondebug:
lgtm with the above changes

@cco3 PTAL

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -57,82 +57,170 @@
...
+
+  public static boolean setPwsEndpoint(Context context, PwsClient pwsClient) {
+    PwsEndpoint endpoint = getCurrentPwsEndpoint(context);
+    pwsClient.setEndpoint(endpoint.url, endpoint.apiVersion, endpoint.apiKey);
+    return endpoint.neededApiKey;
   }
 
-  public static void setPwsEndPointToGoogle(Context context, PwsClient pwsClient){
-    pwsClient.setEndpoint(GOOGLE_ENDPOINT, GOOGLE_ENDPOINT_VERSION, getApiKey(context));
+  public static boolean setPwsEndpoint(Context context,
```
cco3: Document the return values on these...I think it'd make more sense to return whether it was successful (the opposite of what you are returning).

cco3:
Just one comment.  Other than that, LGTM.

hayesjordan:
Updated.

PR #691, opened 2016-06-16
Icon Cache

nishadg246:
Write Icons to our cache instead of
fetching them each time we resume


cco3:
Please put permissions stuff in its own change.  Also, please make the commit message imperative.  e.g., "Cache icons in PhysicalWebCollection"

PR #697, opened 2016-06-21
RSSI Crash

nishadg246:
After retrieving the cache, since we have not cached the RegionResolver
we have to insert the Txpower and Rssi data for each UrlDevice.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
@@ -184,6 +184,12 @@ public class UrlDeviceDiscoveryService extends Service
       JSONObject serializedCollection = new JSONObject(prefs.getString(PW_COLLECTION_KEY, null));
       mPwCollection = PhysicalWebCollection.jsonDeserialize(serializedCollection);
       Utils.setPwsEndpoint(this, mPwCollection);
+      // replace TxPower and RSSI data after restoring cache
+      for (UrlDevice urlDevice : mPwCollection.getUrlDevices()) {
+        if (Utils.isBleUrlDevice(urlDevice)) {
+          Utils.updateRegion(urlDevice);
+        }
+      }
     } catch (JSONException e) {
       Log.e(TAG, "Could not restore Physical Web collection cache", e);
     } catch (PhysicalWebCollectionException e) {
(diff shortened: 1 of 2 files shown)


cco3:
Imperative commit subjects
72 is line length for the message, so you can extend those lines a bit more.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -184,6 +184,9 @@ private void restoreCache() {
       JSONObject serializedCollection = new JSONObject(prefs.getString(PW_COLLECTION_KEY, null));
       mPwCollection = PhysicalWebCollection.jsonDeserialize(serializedCollection);
       Utils.setPwsEndpoint(this, mPwCollection);
+      for (UrlDevice urlDevice : mPwCollection.getDeviceIdToUrlDeviceMap().values()) {
+        Utils.updateRegion(urlDevice);
```
cco3: You need to make sure the URL device has rssi and txpower before doing this (check isBleUrlDevice)

java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -116,6 +116,10 @@ public PwsResult getMetadataByBroadcastUrl(String broadcastUrl) {
     return mBroadcastUrlToPwsResultMap.get(broadcastUrl);
   }
 
+  public Map<String, UrlDevice> getDeviceIdToUrlDeviceMap() {
```
cco3: I'd rather a getUrlDevices() method that returned a List.

nishadg246:
@cco3 fixed

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -184,6 +184,11 @@ private void restoreCache() {
       JSONObject serializedCollection = new JSONObject(prefs.getString(PW_COLLECTION_KEY, null));
       mPwCollection = PhysicalWebCollection.jsonDeserialize(serializedCollection);
       Utils.setPwsEndpoint(this, mPwCollection);
+      for (UrlDevice urlDevice : mPwCollection.getUrlDevices()) {
```
nondebug: lgtm, but could you add a comment explaining why this is necessary?

PR #701, opened 2016-06-22
Fix ListView Adapter Crash

nishadg246:
When the ListView adapter was modified on the background thread instead
of the UI thread, the application would crash.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -259,21 +259,26 @@ public void onListItemClick(ListView l, View v, int position, long id) {
...
-      Log.d(TAG, "groupid to add " + groupId);
-      if (mNearbyDeviceAdapter.containsGroupId(groupId)) {
-        mNearbyDeviceAdapter.updateItem(pwPair);
-      } else if (!mGroupIdQueue.contains(groupId)) {
-        mGroupIdQueue.add(groupId);
-        if (mSecondScanComplete) {
-          // If we've already waited for the second scan timeout, go ahead and put the item in the
-          // listview.
-          emptyGroupIdQueue();
+    new Handler(Looper.getMainLooper()).post(new Runnable() {
```
cco3: Please add a comment here about why this is necessary.  Something like: "The discovery service gives us callbacks on a thread other than the main thread, but we need to update our displayable data on the main thread"

nondebug:
lgtm

PR #707, opened 2016-06-23
Remove getRank

nishadg246:
Before adding distance sorting, the sorting was based on a rank which
was always a value of .5.


java/libs/src/main/java/org/physical_web/collection/PhysicalWebCollection.java
```diff
@@ -324,7 +324,7 @@ public PwPair getTopRankedPwPairByGroupId(String groupId, Comparator<PwPair> com
         // Create the group if it doesn't exist
         UrlGroup urlGroup = groupMap.get(groupId);
         if (urlGroup == null) {
-          urlGroup = new UrlGroup(groupId);
+          urlGroup = new UrlGroup(groupId, comparator);
```
cco3: I don't think the group itself should know about the comparator.  Just pass it as a param to getTopPair

java/libs/src/test/java/org/physical_web/collection/PhysicalWebCollectionTest.java
```diff
@@ -145,46 +129,4 @@ public void jsonDeserializeWorks() throws PhysicalWebCollectionException {
     assertEquals(pwsResult.getSiteUrl(), URL1);
     assertEquals(pwsResult.getGroupId(), GROUP_ID1);
   }
-
-  @Test
-  public void getPwPairsSortedByRankWorks() {
```
cco3: This method still exists and needs to be tested, right?

cco3:
LGTM.  Test the app to make sure it still works right and sorts by distance.

Please add a little more about why this change is desirable (lets client code provide relative sort comparators instead of demanding they provide an absolute rank.)

java/libs/src/test/java/org/physical_web/collection/PwPairTest.java
```diff
@@ -40,7 +40,7 @@
   private static final double RANK2 = 0.9d;
   private UrlDevice mUrlDevice1;
   private PwsResult mPwsResult1;
-  private PwPair mPwPair1;
+  private PwPair mPwPair1 = null;
```
nondebug: This isn't needed as mPwPair1 will already be initialized to null.

PR #711, opened 2016-06-27
Add mDNS scan setting

hayesjordan:


android/PhysicalWeb/app/src/main/res/values/strings.xml
```diff
@@ -49,6 +49,11 @@
     <string name="prod_pws">Production Open Source Physical Web Service</string>
     <string name="dev_pws">Development Open Source Physical Web Service</string>
     <string name="custom_pws">Custom Physical Web Service</string>
+    <string name="scan_settings_key">scan_settings</string>
```
cco3: Only human readable strings should go in here...just make this a private static value

cco3: Ah, nm, I see that the preference xml uses it...it'd be cool if you could make a follow up change to move out all such strings into their own file.

PR #712, opened 2016-06-27
Refactor string resources

hayesjordan:
Move all non-human readable strings to a separate file

### DIFF
--- a/android/PhysicalWeb/app/src/main/AndroidManifest.xml
+++ b/android/PhysicalWeb/app/src/main/AndroidManifest.xml
@@ -89,7 +89,7 @@
 
         <activity
             android:name=".OobActivity"
-            android:label="@string/title_activity_oob"
+            android:label="OobActivity"
             android:screenOrientation="portrait"
             android:theme="@style/Theme.NoActionBar">
         </activity>
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/OobActivity.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/OobActivity.java
@@ -17,9 +17,7 @@
 package org.physical_web.physicalweb;
 
 import android.annotation.SuppressLint;
-import android.content.Context;
 import android.content.Intent;
-import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.support.v7.app.AppCompatActivity;
 import android.view.Menu;
@@ -37,12 +35,7 @@ public class OobActivity extends AppCompatActivity {
     @Override
     public void onClick(View v) {
       // Save the opt in preference
-      String preferencesKey = getString(R.string.main_prefs_key);
-      SharedPreferences sharedPreferences = getSharedPreferences(preferencesKey,
-                                                                 Context.MODE_PRIVATE);
-      SharedPreferences.Editor editor = sharedPreferences.edit();
-      editor.putBoolean(getString(R.string.user_opted_in_flag), true);
-      editor.apply();
+      Utils.setOptInPreference(OobActivity.this);
 
       // Exit the activity
       finish();
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
@@ -27,7 +27,6 @@ import org.physical_web.collection.UrlDevice;
 import android.app.Notification;
 import android.app.PendingIntent;
 import android.app.Service;
-import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.res.Resources;
@@ -36,6 +35,7 @@ import android.os.Binder;
 import android.os.Build;
 import android.os.Handler;
 import android.os.IBinder;
+import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.NotificationManagerCompat;
 import android.util.Log;
@@ -162,8 +162,7 @@ public class UrlDeviceDiscoveryService extends Service
 
   private void restoreCache() {
     // Make sure we are trying to load the right version of the cache
-    String preferencesKey = getString(R.string.discovery_service_prefs_key);
-    SharedPreferences prefs = getSharedPreferences(preferencesKey, Context.MODE_PRIVATE);
+    SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
     int prefsVersion = prefs.getInt(PREFS_VERSION_KEY, 0);
     long now = new Date().getTime();
     if (prefsVersion != PREFS_VERSION) {
@@ -241,13 +240,11 @@ public class UrlDeviceDiscoveryService extends Service
 
   private void saveCache() {
     // Write the PW Collection
-    String preferencesKey = getString(R.string.discovery_service_prefs_key);
-    SharedPreferences prefs = getSharedPreferences(preferencesKey, Context.MODE_PRIVATE);
-    SharedPreferences.Editor editor = prefs.edit();
-    editor.putInt(PREFS_VERSION_KEY, PREFS_VERSION);
-    editor.putLong(SCAN_START_TIME_KEY, mScanStartTime);
-    editor.putString(PW_COLLECTION_KEY, mPwCollection.jsonSerialize().toString());
-    editor.apply();
+    PreferenceManager.getDefaultSharedPreferences(this).edit()
+        .putInt(PREFS_VERSION_KEY, PREFS_VERSION)
+        .putLong(SCAN_START_TIME_KEY, mScanStartTime)
+        .putString(PW_COLLECTION_KEY, mPwCollection.jsonSerialize().toString())
+        .apply();
   }
 
   @Override
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
@@ -27,7 +27,6 @@ import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.content.ServiceConnection;
-import android.content.SharedPreferences;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.net.Uri;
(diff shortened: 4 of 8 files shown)


android/PhysicalWeb/app/src/main/res/values/config-strings.xml
```diff
@@ -0,0 +1,15 @@
...
+<resources>
+    <string name="url_getting_started">https://google.github.io/physical-web/mobile/android/getting-started.html</string>
+    <string name="pws_endpoint_setting_key">pws_endpoint</string>
+    <string name="custom_pws_endpoint_key">custom_pws_endpoint</string>
+    <string name="custom_pws_url_key">custom_pws_url</string>
+    <string name="custom_pws_version_key">custom_pws_version</string>
+    <string name="custom_pws_api_key_key">custom_pws_key</string>
+    <string name="scan_settings_key">scan_settings</string>
+    <string name="mDNS_key">mDNS</string>
+    <string name="main_prefs_key">physical_web_preferences</string>
```
cco3: I don't think this (or it's comment) are needed here...make it a static variable in the class where it's needed.

android/PhysicalWeb/app/src/main/res/values/config-strings.xml
```diff
@@ -0,0 +1,15 @@
...
+    <!-- The following is a more appropriate name, but would require special code to migrate users -->
+    <!-- string name="main_prefs_key">org.physical_web.physicalweb.MAIN_PREFS</string -->
+    <string name="discovery_service_prefs_key">org.physical_web.physicalweb.DISCOVERY_SERVICE_PREFS</string>
```
cco3: Why does this have namespaced name?

hayesjordan: I believe because it is being used for a shared preference file and those need to be unique among all apps.

android/PhysicalWeb/app/src/main/res/values/config-strings.xml
hayesjordan: It is used in 3 different classes but I could put it in Utils?

android/PhysicalWeb/app/src/main/res/values/config-strings.xml
cco3: OK, please just make it a static variable.

android/PhysicalWeb/app/src/main/res/values/config-strings.xml
```diff
@@ -0,0 +1,15 @@
...
+    <string name="url_getting_started">https://google.github.io/physical-web/mobile/android/getting-started.html</string>
+    <string name="pws_endpoint_setting_key">pws_endpoint</string>
+    <string name="custom_pws_endpoint_key">custom_pws_endpoint</string>
+    <string name="custom_pws_url_key">custom_pws_url</string>
+    <string name="custom_pws_version_key">custom_pws_version</string>
+    <string name="custom_pws_api_key_key">custom_pws_key</string>
+    <string name="scan_settings_key">scan_settings</string>
+    <string name="mDNS_key">mDNS</string>
+    <string name="main_prefs_key">physical_web_preferences</string>
+    <!-- The following is a more appropriate name, but would require special code to migrate users -->
```
cco3: Remove the comment...it's no longer applicable

android/PhysicalWeb/app/src/main/res/values/config-strings.xml
cco3: For any of those that makes sense, just use the default preferences...that may get you down to just one file (or even no files)

hayesjordan:
Updated

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/OobActivity.java
```diff
@@ -37,12 +35,7 @@
     @Override
     public void onClick(View v) {
       // Save the opt in preference
-      String preferencesKey = getString(R.string.main_prefs_key);
-      SharedPreferences sharedPreferences = getSharedPreferences(preferencesKey,
-                                                                 Context.MODE_PRIVATE);
-      SharedPreferences.Editor editor = sharedPreferences.edit();
-      editor.putBoolean(getString(R.string.user_opted_in_flag), true);
-      editor.apply();
+      Utils.setOptInPreference(v.getContext());
```
cco3: Why not use `this` as the parameter?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -241,13 +240,12 @@ private void cancelNotifications() {
...
   private void saveCache() {
     // Write the PW Collection
-    String preferencesKey = getString(R.string.discovery_service_prefs_key);
-    SharedPreferences prefs = getSharedPreferences(preferencesKey, Context.MODE_PRIVATE);
-    SharedPreferences.Editor editor = prefs.edit();
-    editor.putInt(PREFS_VERSION_KEY, PREFS_VERSION);
-    editor.putLong(SCAN_START_TIME_KEY, mScanStartTime);
-    editor.putString(PW_COLLECTION_KEY, mPwCollection.jsonSerialize().toString());
-    editor.apply();
+    SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
```
cco3: just put this on one line: PreferenceManager.getDefaultSharedPreferences(this).edit() (etc)

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -137,6 +141,10 @@ private static void throwEncodeException(JSONException e) {
     throw new RuntimeException("Could not encode JSON", e);
   }
 
+  private static void deleteFile(Context context, String fileName) {
+    context.deleteFile(fileName);
```
cco3: Use prefs.edit().clear().apply()

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -190,6 +198,21 @@ private static String readCustomPwsEndpointApiKey(Context context) {
...
+        Context.MODE_PRIVATE);
+    return sharedPreferences.getBoolean(USER_OPTED_IN_KEY, false);
+  }
+  /**
+   * Gets whether the user has optedIn from SharedPreferences.
+   * @param context The context for the SharedPreferences.
+   * @return True if the user has optedIn otherwise false.
+   */
+  public static boolean checkIfUserHasOptedIn(Context context) {
+    SharedPreferences sharedPref = PreferenceManager.getDefaultSharedPreferences(context);
```
cco3: Just make this a one-liner

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -190,6 +198,21 @@ private static String readCustomPwsEndpointApiKey(Context context) {
     return sharedPref.getString(context.getString(R.string.custom_pws_api_key_key), "");
   }
 
+  private static boolean restoreMainPrefSettings(Context context) {
```
cco3: I don't think this should be its own method...just copy this code down to the method where you use it.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -241,6 +273,11 @@ public static void setPwsEndpointPreference(Context context, String endpoint) {
   public static void setSharedPreferencesDefaultValues(Context context) {
     PreferenceManager.setDefaultValues(context, R.xml.settings, false);
     setPwsEndpointPreference(context, getCurrentPwsEndpointString(context));
+    if (restoreMainPrefSettings(context)) {
+      setOptInPreference(context);
+      deleteFile(context, MAIN_PREFS_KEY);
```
cco3: Remind me: this is the cache and what else?

android/PhysicalWeb/app/src/main/res/values/strings.xml
```diff
@@ -62,11 +54,6 @@
     <string name="title_activity_oob">OobActivity</string>
```
cco3: Should this be here?  If it's supposed to be for humans to read, maybe we should change the string.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/OobActivity.java
hayesjordan: I tried that initially but in `this` refers to the `onClickListener`

cco3: use `OobActivity.this`

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
hayesjordan: I was under the impression that didn't delete the file but only cleared the contents. Does Android automatically delete empty preference files?

cco3: I don't know, but it's probably not all that important...are you sure the filename you are using is even the right filename here?  If you have to manually determine the file name, then it's probably not the right way of doing it.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
hayesjordan: The `MAIN_PREFS_KEY` refers to the opt in and the `DISCOVERY_SERVICE_PREFS_KEY` refers to the cache.

android/PhysicalWeb/app/src/main/res/values/strings.xml
hayesjordan: It doesn't appear to be used for anything the user sees. I will just move it to the other file.

cco3: Can it just be a static value then?

hayesjordan: Its not used in any java files, it is only used in the AndroidManifest. Can I just place the literal there?

hayesjordan:
Updated

android/PhysicalWeb/app/src/main/AndroidManifest.xml
```diff
@@ -89,7 +89,7 @@
 
         <activity
             android:name=".OobActivity"
-            android:label="@string/title_activity_oob"
+            android:label="OobActivity"
```
cco3: Ah, I see...you could have kept it if it's config to config... no worries though.  Just leave it.

cco3:
LGTM

PR #715, opened 2016-06-30
Added Swipe to Hide

nishadg246:
Swipe away URLs to make it easier to browse nearby beacons


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -202,6 +202,23 @@ private void initialize(View rootView) {
...
+                  }
+
+                  @Override
+                  public void onDismiss(ListView listView, int position) {
+                  }
+              });
+    listView.setOnTouchListener(touchListener);
+    // Setting this scroll listener is required to ensure that during ListView scrolling,
+    // we don't look for swipes.
+    listView.setOnScrollListener(touchListener.makeScrollListener());
```
cco3: let's comment this out for now

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -265,22 +282,24 @@ public void onUrlDeviceDiscoveryUpdate() {
...
-          Log.d(TAG, "groupid to add " + groupId);
-          if (mNearbyDeviceAdapter.containsGroupId(groupId)) {
-            mNearbyDeviceAdapter.updateItem(pwPair);
-          } else if (!mGroupIdQueue.contains(groupId)) {
-            mGroupIdQueue.add(groupId);
-            if (mSecondScanComplete) {
-              // If we've already waited for the second scan timeout,
-              // go ahead and put the item in the listview.
-              emptyGroupIdQueue();
+        if (!SwipeDismissListViewTouchListener.isLocked()) {
```
cco3: just return if it's locked...makes this easier to read

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -333,18 +352,20 @@ public void onRefresh() {
   }
 
   private void emptyGroupIdQueue() {
-    List<PwPair> pwPairs = new ArrayList<>();
+    if (!SwipeDismissListViewTouchListener.isLocked()) {
+      List<PwPair> pwPairs = new ArrayList<>();
```
cco3: hm, you can't just not empty the queue if it's locked...you'll have to empty it later.  Maybe you can empty it and just not call notifyDatasetchanged (and then call notifyDataSetChanged in onDismiss).  The same applies above for adding new urls.

PR #721, opened 2016-07-07
Add revised mDNS scanning

nishadg246:
Allows for broadcasting of public or private URLs over mDNS.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
@@ -144,6 +144,7 @@ public class UrlDeviceDiscoveryService extends Service
     mUrlDeviceDiscoverers = new ArrayList<>();
 
     if (Utils.getMdnsEnabled(this)) {
+      Log.d(TAG,"mdns started");
       mUrlDeviceDiscoverers.add(new MdnsUrlDeviceDiscoverer(this));
     }
     mUrlDeviceDiscoverers.add(new SsdpUrlDeviceDiscoverer(this));
@@ -232,6 +233,7 @@ public class UrlDeviceDiscoveryService extends Service
   public void onRebind(Intent intent) {
     mIsBound = true;
   }
+
   private void cancelNotifications() {
     mNotificationManager.cancel(NEAREST_BEACON_NOTIFICATION_ID);
     mNotificationManager.cancel(SECOND_NEAREST_BEACON_NOTIFICATION_ID);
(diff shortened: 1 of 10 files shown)


cco3:
"protocol" is a bit of a misnomer...maybe just say "scanning" or "format"...those are the best words I can think of at the moment.

Also, please use imperative mood.  ("Add")

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -268,6 +268,13 @@ public void onUrlDeviceDiscovered(UrlDevice urlDevice) {
     // Don't short circuit because icons
     // and metadata may not be fetched
     mPwCollection.addUrlDevice(urlDevice);
+    if (!Utils.isPublic(urlDevice)) {
+      mPwCollection.addMetadata(new PwsResult.Builder(urlDevice.getUrl(), urlDevice.getUrl())
+        .setTitle(Utils.getTitle(urlDevice))
+        .setDescription(Utils.getDescription(urlDevice))
+        .build());
+      Log.d("done","done2");
```
cco3: remove

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/UrlDeviceDiscoveryService.java
```diff
@@ -268,6 +268,13 @@ public void onUrlDeviceDiscovered(UrlDevice urlDevice) {
     // Don't short circuit because icons
     // and metadata may not be fetched
     mPwCollection.addUrlDevice(urlDevice);
+    if (!Utils.isPublic(urlDevice)) {
```
cco3: "isPublic" is currently used to determine whether or not the result should be private to the user (for example, would it be OK to show on the lock screen if they don't want private data displayed on the lock screen).  This is different from the concept of a URL being available only on the local network.  Instead, just make an isLocalDevice() function that checks for title & description.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -567,6 +600,24 @@ public UrlDeviceBuilder setPublic() {
     }
 
     /**
+     * Set the title
+     * @return The builder with title
+     */
+    public UrlDeviceBuilder setTitle(String title) {
+      addExtra(TITLE_KEY, title);
```
cco3: Just do return addExtra().  Fix everywhere else that doesn't.

PR #724, opened 2016-07-15
Add Favoriting and Blocking Functionality

nishadg246:
Added the ability to favorite URLs to get higher level notifications from and the ability to block URLs from certain domains.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -78,6 +78,7 @@
   private boolean mSecondScanComplete;
   private boolean mFirstTime;
   private DiscoveryServiceConnection mDiscoveryServiceConnection;
+  private boolean missedEmptyGroupIdQueue = false;
```
cco3: * mMissed
* set this elsewhere (in onCreate I guess)

cco3:
Swipe to hide looks OK

PR #725, opened 2016-07-15
Fix ListView Click

nishadg246:
Clicking a URL would not send the intent until scanning was completed.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
@@ -94,6 +94,7 @@ public class NearbyBeaconsFragment extends ListFragment
       if (!mGroupIdQueue.isEmpty()) {
         emptyGroupIdQueue();
         mSwipeRefreshWidget.setRefreshing(false);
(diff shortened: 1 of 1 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -107,6 +108,7 @@ public void run() {
       mSecondScanComplete = true;
       mSwipeRefreshWidget.setRefreshing(false);
       mScanningAnimationTextView.setAlpha(0f);
```
cco3: What is this needed for?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -308,6 +310,7 @@ private void startScanningDisplay(long scanStartTime, boolean hasResults) {
       mScanningAnimationDrawable.start();
     } else {
       mSwipeRefreshWidget.setRefreshing(false);
+      mScanningAnimationDrawable.stop();
```
cco3: In what circumstances is this needed?

PR #733, opened 2016-07-20
Fix URL not clickable

hayesjordan:


cco3:
What exactly was the new descendant getting in the way?

cco3:
Also, will this break the star functionality?

hayesjordan:
When the star functionality was added it took away the ability to click the URLs because it's click handler was disabling the list item click handler. When you block focusability it doesn't disable the item's click handler.

hayesjordan:
This does not break the star functionality, but may not work with certain hardware. 

> Note: Because of this trick, your layout no longer works using a directional pad or a trackball. Indeed, we cannot select a accessories of an itemview. I don’t consider this as a major problem because most Android devices are now based on a touch-screen and the Android Market automatically adds a hardware requirement for a touchscreen to all applications (android.hardware.touchscreen). If you want your code to be safe, you can add a `<uses-hardware />` tag in your AndroidManifest.xml

cco3:
LGTM

PR #734, opened 2016-07-20
Added revised mDNS protocol

nishadg246:
Allows for broadcasting of public or private URLs over mDNS.

### DIFF
--- a/android/PhysicalWeb/app/build.gradle
+++ b/android/PhysicalWeb/app/build.gradle
@@ -75,6 +75,7 @@ dependencies {
     compile 'com.android.volley:volley@aar'
     compile 'org.uribeacon:uribeacon-library-release@aar'
     compile 'com.android.support:appcompat-v7:23.4.0'
+    compile 'com.youview:tinydnssd:1.2.7'
 }
 
 checkstyle {
--- a/android/PhysicalWeb/app/src/main/AndroidManifest.xml
+++ b/android/PhysicalWeb/app/src/main/AndroidManifest.xml
@@ -12,6 +12,12 @@
     <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
+    <uses-permission
+        android:required="true"
+        android:name="android.permission.ACCESS_WIFI_STATE"/>
+    <uses-permission
+        android:required="true"
+        android:name="android.permission.CHANGE_WIFI_STATE"/>
     <!-- This is required for the scan library. -->
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
 
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
@@ -17,74 +17,59 @@
 package org.physical_web.physicalweb;
 
 import android.content.Context;
-import android.net.nsd.NsdManager;
-import android.net.nsd.NsdServiceInfo;
 import android.util.Log;
 import android.webkit.URLUtil;
 
+import com.youview.tinydnssd.DiscoverResolver;
+import com.youview.tinydnssd.MDNSDiscover;
+
+import java.util.Map;
+
+
 class MdnsUrlDeviceDiscoverer extends UrlDeviceDiscoverer {
-  private static final String TAG = "MdnsUrlDeviceDiscoverer";
-  NsdManager.DiscoveryListener mDiscoveryListener = new NsdManager.DiscoveryListener() {
-
-    @Override
-    public void onDiscoveryStarted(String regType) {
-      Log.d(TAG, "Service discovery started");
-      mState = State.STARTED;
-    }
-
-    @Override
-    public void onServiceFound(NsdServiceInfo service) {
-      Log.d(TAG, "Service discovery success" + service);
-      String name = service.getServiceName();
-      if (URLUtil.isNetworkUrl(name)) {
-        String id = TAG + service.getHost() + service.getPort();
-        reportUrlDevice(createUrlDeviceBuilder(id, name)
-            .setPrivate()
-            .build());
-      }
-    }
-
-    @Override
-    public void onServiceLost(NsdServiceInfo service) {
-      Log.e(TAG, "service lost" + service);
-    }
-
-    @Override
-    public void onDiscoveryStopped(String serviceType) {
-      Log.i(TAG, "Discovery stopped: " + serviceType);
-      mState = State.STOPPED;
-      if (toRestart) {
-        toRestart = false;
-        startScan();
-      }
-    }
-
-    @Override
-    public void onStartDiscoveryFailed(String serviceType, int errorCode) {
-      Log.e(TAG, "Discovery failed: Error code:" + errorCode);
-      mNsdManager.stopServiceDiscovery(this);
-    }
-
-    @Override
-    public void onStopDiscoveryFailed(String serviceType, int errorCode) {
-      Log.e(TAG, "Discovery failed: Error code:" + errorCode);
-      mNsdManager.stopServiceDiscovery(this);
-    }
-  };
-  private static final String MDNS_SERVICE_TYPE = "_http._tcp.";
-  private NsdManager mNsdManager;
+  private static final String TAG = "mdns";
+  private DiscoverResolver mResolver;
+  private static final String MDNS_SERVICE_TYPE = "_physicalweb._tcp";
   private enum State {
     STOPPED,
     WAITING,
     STARTED,
   }
   private State mState;
-  private boolean toRestart;
 
   public MdnsUrlDeviceDiscoverer(Context context) {
-    mNsdManager = (NsdManager) context.getSystemService(Context.NSD_SERVICE);
     mState = State.STOPPED;
-    toRestart = false;
+    resolver = new DiscoverResolver(context, MDNS_SERVICE_TYPE,
+        new DiscoverResolver.Listener() {
+      @Override
+      public void onServicesChanged(Map<String, MDNSDiscover.Result> services) {
+        for (MDNSDiscover.Result result : services.values()) {
+          // access the Bluetooth MAC from the TXT record
+          String url = result.txt.dict.get("url");
+          Log.d(TAG, url);
+          String id = TAG + result.srv.fqdn + result.srv.port;
+          String title = "";
+          String description = "";
+          if ("false".equals(result.txt.dict.get("public"))) {
+            if (result.txt.dict.containsKey("title")) {
+              title = result.txt.dict.get("title");
+            }
+            if (result.txt.dict.containsKey("description")) {
+              description = result.txt.dict.get("description");
+            }
+            reportUrlDevice(createUrlDeviceBuilder(id, url)
+              .setPrivate()
+              .setTitle(title)
+              .setDescription(description)
+              .build());
+          } else if (URLUtil.isNetworkUrl(url)) {
+            reportUrlDevice(createUrlDeviceBuilder(id, url)
+              .setPrivate()
+              .build());
+          }
+        }
+      }
+    });
   }
 
   @Override
@@ -93,7 +78,7 @@ class MdnsUrlDeviceDiscoverer extends UrlDeviceDiscoverer {
       return;
     }
     mState = State.WAITING;
-    mNsdManager.discoverServices(MDNS_SERVICE_TYPE, NsdManager.PROTOCOL_DNS_SD, mDiscoveryListener);
+    resolver.start();
   }
 
   @Override
@@ -102,12 +87,6 @@ class MdnsUrlDeviceDiscoverer extends UrlDeviceDiscoverer {
       return;
     }
     mState = State.WAITING;
-    mNsdManager.stopServiceDiscovery(mDiscoveryListener);
-  }
-
-  @Override
-  public synchronized void restartScan() {
-    toRestart = true;
-    stopScan();
+    resolver.stop();
   }
 }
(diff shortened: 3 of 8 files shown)


android/PhysicalWeb/app/build.gradle
```diff
@@ -75,6 +75,7 @@ dependencies {
     compile 'com.android.volley:volley@aar'
     compile 'org.uribeacon:uribeacon-library-release@aar'
     compile 'com.android.support:appcompat-v7:23.4.0'
+    compile 'com.youview:tinydnssd:1.2.7'
```
cco3: Do we just want to use the standard library and limit this functionality to N?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -344,21 +348,39 @@ private void emptyGroupIdQueue() {
...
     NearbyBeaconsAdapter() {
       mPwPairs = new ArrayList<>();
+      numberOfHideableResults = 0;
     }
 
     public void addItem(PwPair pwPair) {
-      mPwPairs.add(pwPair);
+      if (Utils.isLocalDevice(pwPair.getUrlDevice())) {
+        if (numberOfHideableResults == 0) {
+          mPwPairs.add(new PwPair(null, new PwsResult(null, null)));
```
cco3: Add a comment explaining this

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -67,7 +67,9 @@
   private static final String DISCOVERY_SERVICE_PREFS_KEY =
       "org.physical_web.physicalweb.DISCOVERY_SERVICE_PREFS";
   private static final String SCANTIME_KEY = "scantime";
-  private static final String PUBLIC_KEY = "public";
+  public static final String PUBLIC_KEY = "public";
```
cco3: are these supposed to be public?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -549,6 +551,20 @@ public static boolean isBleUrlDevice(UrlDevice urlDevice) {
   }
 
   /**
+   * Checks if device is Local.
+   * @param urlDevice The device that is getting checked.
+   * @return If the device is local or not.
+   */
+  public static boolean isLocalDevice(UrlDevice urlDevice) {
```
cco3: Something other than local would be better...isResolvable might make more sense

nishadg246:
@cco3 fixed

cco3:
LGTM from this and previous versions of the change

nishadg246:
@nondebug

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
```diff
@@ -17,74 +17,63 @@
...
-    public void onStopDiscoveryFailed(String serviceType, int errorCode) {
-      Log.e(TAG, "Discovery failed: Error code:" + errorCode);
-      mNsdManager.stopServiceDiscovery(this);
-    }
-  };
-  private static final String MDNS_SERVICE_TYPE = "_http._tcp.";
-  private NsdManager mNsdManager;
+class MdnsUrlDeviceDiscoverer extends UrlDeviceDiscoverer {
+  private static final String TAG = "mdns";
+  DiscoverResolver resolver;
```
nondebug: private DiscoverResolver mResolver

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
```diff
@@ -17,74 +17,63 @@
...
-    mNsdManager = (NsdManager) context.getSystemService(Context.NSD_SERVICE);
     mState = State.STOPPED;
-    toRestart = false;
+    resolver = new DiscoverResolver(context, MDNS_SERVICE_TYPE,
+    new DiscoverResolver.Listener() {
+      @Override
+      public void onServicesChanged(Map<String, MDNSDiscover.Result> services) {
+        for (MDNSDiscover.Result result : services.values()) {
+            // access the Bluetooth MAC from the TXT record
+            Log.d(TAG, result.txt.dict.toString());
```
nondebug: .toString() is unnecessary here, Log.d(TAG, result.txt.dict) will also print the dict's string representation

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
```diff
@@ -17,74 +17,63 @@
...
+            String url = result.txt.dict.get("url");
+            Log.d(TAG, url);
+            String id = TAG + result.srv.fqdn + result.srv.port;
+            String title = "";
+            String description = "";
```
nondebug: Strings are automatically initialized to the empty string so the assignments are unnecessary here

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
```diff
@@ -17,74 +17,63 @@
...
+            if (result.txt.dict.containsKey("public") &&
+                result.txt.dict.get("public").equals("false")) {
```
nondebug: You could reverse the comparison to avoid calling containsKey:
"false".equals(result.txt.dict.get("public"))
If the key isn't present it will return null, and .equals(null) will return false

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
```diff
@@ -17,74 +17,63 @@
...
+              }
+              if (result.txt.dict.containsKey("description")) {
+                description = result.txt.dict.get("description");
+              }
+              reportUrlDevice(createUrlDeviceBuilder(id, url)
+                .setPrivate()
+                .setTitle(title)
+                .setDescription(description)
+                .build());
+              continue;
```
nondebug: instead of using continue here, could you make the next block "else if"?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -344,21 +348,39 @@ private void emptyGroupIdQueue() {
     mNearbyDeviceAdapter.notifyDataSetChanged();
   }
 
+  private boolean isFolderItem(PwPair item) {
```
nondebug: make this static

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -344,21 +348,39 @@ private void emptyGroupIdQueue() {
...
+    return item.getUrlDevice() == null && item.getPwsResult().getSiteUrl() == null;
+  }
+
   // Adapter for holding beacons found through scanning.
   private class NearbyBeaconsAdapter extends BaseAdapter {
     private List<PwPair> mPwPairs;
+    private int numberOfHideableResults;
```
nondebug: mNumberOfHideableResults

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -398,15 +427,20 @@ private void setText(View view, int textViewId, String text) {
...
       // Display the pwsResult.
       PwPair pwPair = getItem(i);
       PwsResult pwsResult = pwPair.getPwsResult();
+      if (isFolderItem(pwPair)) {
+        view = getActivity().getLayoutInflater().inflate(R.layout.folder_item_nearby_beacon,
+            viewGroup, false);
+        WifiManager wifiManager = (WifiManager) getActivity().
+            getSystemService(Context.WIFI_SERVICE);
+        WifiInfo wifiInfo = wifiManager.getConnectionInfo();
+        setText(view, R.id.title, wifiInfo.getSSID().trim().replace("\"", ""));
```
nondebug: It would be better to only strip off the leading and trailing quotes rather than removing all quotes since there might be " characters in the SSID itself.

android/PhysicalWeb/app/src/main/AndroidManifest.xml
```diff
@@ -12,6 +12,12 @@
     <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
+        <uses-permission
```
nondebug: fix indent

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
```diff
@@ -17,74 +17,63 @@
...
   public MdnsUrlDeviceDiscoverer(Context context) {
-    mNsdManager = (NsdManager) context.getSystemService(Context.NSD_SERVICE);
     mState = State.STOPPED;
-    toRestart = false;
+    resolver = new DiscoverResolver(context, MDNS_SERVICE_TYPE,
+    new DiscoverResolver.Listener() {
+      @Override
+      public void onServicesChanged(Map<String, MDNSDiscover.Result> services) {
+        for (MDNSDiscover.Result result : services.values()) {
+            // access the Bluetooth MAC from the TXT record
```
nondebug: dedent this block one level

nishadg246:
@nondebug fixed

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/MdnsUrlDeviceDiscoverer.java
```diff
@@ -17,74 +17,59 @@
...
-    public void onStopDiscoveryFailed(String serviceType, int errorCode) {
-      Log.e(TAG, "Discovery failed: Error code:" + errorCode);
-      mNsdManager.stopServiceDiscovery(this);
-    }
-  };
-  private static final String MDNS_SERVICE_TYPE = "_http._tcp.";
-  private NsdManager mNsdManager;
+class MdnsUrlDeviceDiscoverer extends UrlDeviceDiscoverer {
+  private static final String TAG = "mdns";
+  private DiscoverResolver resolver;
```
nondebug: mResolver?

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -398,15 +430,21 @@ private void setText(View view, int textViewId, String text) {
...
       PwPair pwPair = getItem(i);
       PwsResult pwsResult = pwPair.getPwsResult();
+      if (isFolderItem(pwPair)) {
+        view = getActivity().getLayoutInflater().inflate(R.layout.folder_item_nearby_beacon,
+            viewGroup, false);
+        WifiManager wifiManager = (WifiManager) getActivity().
+            getSystemService(Context.WIFI_SERVICE);
+        WifiInfo wifiInfo = wifiManager.getConnectionInfo();
+        String ssid = wifiInfo.getSSID().trim();
+        setText(view, R.id.title, ssid.substring(1,ssid.length() -1));
```
nondebug: According to the getSSID documentation, it's possible that the returned SSID won't have the leading and trailing quotes if the SSID can't be decoded as UTF-8:

Returns the service set identifier (SSID) of the current 802.11 network. If the SSID can be decoded as UTF-8, it will be returned surrounded by double quotation marks. Otherwise, it is returned as a string of hex digits. The SSID may be <unknown ssid> if there is no network currently connected.

https://developer.android.com/reference/android/net/wifi/WifiInfo.html#getSSID()

We should detect the hex digits case and do something reasonable.  I think it's enough to display something generic like "Wireless Network".

android/PhysicalWeb/app/src/main/res/layout/list_item_nearby_beacon.xml
```diff
@@ -94,6 +94,8 @@
 
     </LinearLayout>
 
+
```
nondebug: remove the added newlines

nishadg246:
@nondebug fixed

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -398,15 +430,25 @@ private void setText(View view, int textViewId, String text) {
...
       PwsResult pwsResult = pwPair.getPwsResult();
+      if (isFolderItem(pwPair)) {
+        view = getActivity().getLayoutInflater().inflate(R.layout.folder_item_nearby_beacon,
+            viewGroup, false);
+        WifiManager wifiManager = (WifiManager) getActivity().
+            getSystemService(Context.WIFI_SERVICE);
+        WifiInfo wifiInfo = wifiManager.getConnectionInfo();
+        String ssid = wifiInfo.getSSID().trim();
+        if (ssid.charAt(0) == '"' && ssid.charAt(ssid.length()-1) == '"') {
+          setText(view, R.id.title, ssid.substring(1,ssid.length() -1));
```
nondebug: Add spaces before and after arithmetic operators and after commas.
```
        if (ssid.charAt(0) == '"' && ssid.charAt(ssid.length() - 1) == '"') {
          setText(view, R.id.title, ssid.substring(1, ssid.length() - 1));
```
LGTM otherwise

PR #746, opened 2016-07-27
Fix debug view setting

hayesjordan:


android/PhysicalWeb/app/src/main/res/xml/settings.xml
```diff
@@ -46,7 +46,7 @@
         android:persistent="false">
 
         <CheckBoxPreference
-           android:key="@string/debug_view_key"
+           android:key="@string/debug_key"
```
cco3: Where is this defined?

hayesjordan: config-strings.xml

Issue #749, opened 2016-07-28
Prod/Dev endpoints swapped in PWS Settings


nondebug:
When selecting the Production Open Source Physical Web Service, the Development service is used instead (and vice versa).

PR #752, opened 2016-07-29
Fix Pws Ordering in settings list

hayesjordan:
Fix scanning with invalid PWS

Issue #749

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
@@ -278,29 +278,7 @@ class Utils {
 
   }
 
-  private static class ClearCacheDiscoveryServiceConnection implements ServiceConnection {
-    private Context mContext;
 
-    @Override
-    public synchronized void onServiceConnected(ComponentName className, IBinder service) {
-      // Get the service
-      UrlDeviceDiscoveryService.LocalBinder localBinder =
-          (UrlDeviceDiscoveryService.LocalBinder) service;
-      localBinder.getServiceInstance().clearCache();
-      mContext.unbindService(this);
-    }
-
-    @Override
-    public synchronized void onServiceDisconnected(ComponentName className) {
-    }
-
-    public synchronized void connect(Context context) {
-      mContext = context;
-      Intent intent = new Intent(mContext, UrlDeviceDiscoveryService.class);
-      mContext.startService(intent);
-      mContext.bindService(intent, this, Context.BIND_AUTO_CREATE);
-    }
-  }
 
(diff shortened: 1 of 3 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -253,23 +253,46 @@ private boolean needApiKey() {
...
     public synchronized void onServiceDisconnected(ComponentName className) {
     }
 
-    public synchronized void connect(Context context) {
+    public synchronized void clearCache(Context context) {
+      mRequestedAction = ACTION_CLEAR_CACHE;
+      connect(context);
+    }
+
+    public synchronized void startScan(Context context) {
```
cco3: Why not just created a new kind of connection in addition to the ClearCacheDiscoveryServiceConnection?  Then you wouldn't need to do any synchronization.

hayesjordan:
@cco3 Updated

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -351,7 +329,59 @@ public static boolean checkIfUserHasOptedIn(Context context) {
    * @param context The context for the service.
    */
   public static void deleteCache(Context context) {
-    new ClearCacheDiscoveryServiceConnection().connect(context);
+    (new ServiceConnection(){
```
cco3: no parens needed here

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/Utils.java
```diff
@@ -351,7 +329,59 @@ public static boolean checkIfUserHasOptedIn(Context context) {
...
+      }
+    }).connect(context);
+  }
+
+  /**
+   * Starts scanning with UrlDeviceDisoveryService.
+   * @param context The context for the service.
+   */
+  public static void startScan(Context context) {
+    (new ServiceConnection(){
```
cco3: I'm fine with this as is, but I'm wondering if it would be worthwhile to subclass service connection so that any implementing subclass gets a localBinder and gets to do stuff with it and doesn't have to manage the connection.

hayesjordan: I'm not exactly sure what you mean by this?

cco3: Let's talk about it later.  I've assigned you to an issue.

cco3:
LGTM

PR #776, opened 2016-08-09
Add PW Logger

hayesjordan:
Amends PW onto the tag before calling the Android Log


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/PhysicalWebLog.java
```diff
@@ -0,0 +1,74 @@
+package org.physical_web.physicalweb;
+
+import android.util.Log;
+
+/**
+ * Created by hayesjordan on 8/5/16.
+ */
+public class PhysicalWebLog {
```
cco3: We should give this class a short name, like Log.  Just specify `android.util.Log` by its fully qualified name here and rename this class to Log.

hayesjordan:
Updated

cco3:
Thanks!  LGTM

PR #783, opened 2016-08-18
Remove beacon configuration

hayesjordan:


cco3:
LGTM, but let's not put this one in just yet...

edent:
Just wondered if there was a reason for this? Spent half an hour trying to figure out why I couldn't edit a beacon. What should I use instead?

cco3:
Handled in #826

PR #795, opened 2016-08-26
Fix back button in NearbyBeaconsFragment

hayesjordan:
Hitting back button in NearbyBeaconsFragment closed the fragment leaving
an empty view. The back button should close the app in this case.


nondebug:
lgtm

PR #801, opened 2016-08-29
Tweek UI to better show action in demos

hayesjordan:
When a user clicked on a Demo all that happened was a Toast and the notification. Scott wanted there to be a little something more to notify users they clicked on something.


cco3:
LGTM

PR #805, opened 2016-08-30
Fix double downloads and zombie downloads

### LINKED ISSUE #803: Double tapping Fatbeacon results in broken pages
When tapping a Fatbeacon result from the Nearby Beacons fragment, if the user inadvertently taps more than once before the Bluetooth connection is made, the app doubles up the download process resulting in broken page fragments being displayed.

Furthermore, a bug allowed the progress window to be closed/cancelled while leaving the connected download service running.

I'm submitting a PR that fixes both issues for review.

angst7:
Fixes #803.  

There still seems to be an edge case that can cause the app to misbehave if you rapidly and repeatedly tap on a Fatbeacon.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BluetoothSite.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BluetoothSite.java
@@ -55,9 +55,15 @@ public class BluetoothSite extends BluetoothGattCallback {
   private ProgressDialog progress;
   private int transferRate = 20;
   private StringBuilder html;
+  private boolean running = false;
 
   public BluetoothSite(Activity activity) {
     this.activity = activity;
+  }
+
+
+  public Boolean isRunning() {
+    return running;
   }
 
   /**
@@ -67,6 +73,7 @@ public class BluetoothSite extends BluetoothGattCallback {
    * @param title The title of the web page being downloaded
    */
   public void connect(String deviceAddress, String title) {
+    running = true;
     String progressTitle = activity.getString(R.string.page_loading_title) + " " + title;
     progress = new ProgressDialog(activity);
     progress.setCancelable(true);
@@ -80,7 +87,7 @@ public class BluetoothSite extends BluetoothGattCallback {
     progress.setTitle(progressTitle);
     progress.setMessage(activity.getString(R.string.page_loading_message));
     progress.show();
-    BluetoothAdapter.getDefaultAdapter().getRemoteDevice(deviceAddress)
+    mBluetoothGatt = BluetoothAdapter.getDefaultAdapter().getRemoteDevice(deviceAddress)
         .connectGatt(activity, false, this);
   }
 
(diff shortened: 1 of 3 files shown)


angst7:
Committed this with the wrong email.  That email now has a CLA, too.  Let me know if I need to redo the PR or if the bot can re-run its check.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BluetoothSite.java
```diff
@@ -54,18 +54,25 @@
   private ProgressDialog progress;
   private int transferRate = 20;
   private StringBuilder html;
+  private Boolean running = false;
```
cco3: Any reason this isn't a primitive?

cco3:
Thanks for the change!  We'll take a look...

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BluetoothSite.java
angst7: I'd like to blame an overly ambitious shift key.  But yeah, it could be a primitive.

angst7:
The other edge case I describe above should be fixed with this followup patch.  What happened was if the second and third taps come in before the connection as established, then:
- The second tap cancelled the dialog and set running to false, but there was no gatt handle to close the connection
- The third tap started a new connection, and showed a new progress while the first connection was still alive.

The issue was simply that mBluetoothGatt was not being set on .connect(), setting it fixes that. 

The other edge case would cause the progress dialog to remain open after an aborted connection.  The onClientStateChanged() method was trapping the disconnect event, but not clearing the progress dialog, so I added the close() method to fix that.

Interestingly I'm not sure what causes this particular problem  It seems to happen when you click on a fatbeacon result quicky after its discovered, while the scan process is still running.  The connection is made, but quickly followed by a status == 133 and a disconnect.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -285,7 +287,9 @@ public void onListItemClick(ListView l, View v, int position, long id) {
       // Initiate WifiDirect Connection request to device
       mWifiDirectConnect.connect(item.getUrlDevice(), pwsResult.getTitle());
     } else if (Utils.isFatBeaconDevice(item.getUrlDevice())) {
-      (new BluetoothSite(getActivity())).connect(pwsResult.getSiteUrl(), pwsResult.getTitle());
+      if (!mBluetoothSite.isRunning()) {
```
hayesjordan: Can this be added to else if above? 
`else if (Utils.isFatBeaconDevice(item.getUrlDevice() && !mBluetoothSite.isRunning()) {`

hayesjordan:
@angst7 

Once you fix the merge conflict and address my comment, can you squash this into one commit?

The git command is `git rebase -i HEAD~2` or 3 if you use another commit to address the comments.

Thanks for fixing the bugs!

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
angst7: No, that would cause the final else case to trigger a createNavigateToUrl intent.

hayesjordan: Didn't think that one through, my mistake.

angst7:
Ok, I think it's all cleaned up and ready.

PR #812, opened 2016-08-31
Fix hanging when connecting to FatBeacon

hayesjordan:
Some devices cannot connect to Gatt service and will hang on the "Please
wait while page loads" dialog. This seams to be because some devices
need to run Ble interactions on the UI thread.


hayesjordan:
This is the post that states that some devices need to run the Ble interactions on the UI thread.
http://stackoverflow.com/questions/28894111/android-ble-gatt-error133-on-connecting-to-device

nondebug:
lgtm

PR #818, opened 2016-09-06
Added the exceptions on jsons in PhysicalWebCollection

RajPratim21:


RajPratim21:
During the initial setup src/java files can't find the files in the folder lib: so I also moved pwsCollections from lib/org.pws.collections to app/main/org.physical.web/collections

cco3:
Are you  trying to develop in android studio? We've had trouble getting the library set up correctly in studio.  Just build from the command line, e.g., `adb installDebug`

Anyway, you are going to have to put the library back where it's supposed to be.

RajPratim21:
Yes I was building with android studio

RajPratim21:
I have a query It seems to  works with the libs under the src\main in my system can't it be done like this? I am a novice so I am apologizing for the naive comments though.

cco3:
We have this code separated out into a distinct library so that it can be used by itself.  We don't want to move that library under the directory for the app.

cco3:
Closing due to inactivity.  Please let me know if you wish to pursue this.

PR #830, opened 2016-09-19
Fix findbugs errors

cco3:


android/PhysicalWeb/app/src/main/java/org/physical_web/demos/FatBeaconHelloWorld.java
```diff
@@ -27,7 +27,7 @@
  */
 public class FatBeaconHelloWorld implements Demo {
   private static final String TAG = FatBeaconHelloWorld.class.getSimpleName();
-  private static boolean mIsDemoStarted = false;
+  private boolean mIsDemoStarted = false;
```
hayesjordan: I think this needs to be static because it needs to retain that state even when the user goes back to the list.

cco3:
Updated

hayesjordan:
LGTM

PR #832, opened 2016-09-20
Reorganize BluetoothSite logic

cco3:
Instead of:
1. Converting bytes to a string,
2. Adding to a string builder,
3. Converting back to bytes, and
4. Writing bytes to a file

we just write the bytes directly to a file.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/BluetoothSite.java
```diff
@@ -187,6 +204,16 @@ private void close() {
       progress.dismiss();
     }
 
+    if (mHtml != null) {
+      try {
+        mHtml.close();
+      } catch (IOException e) {
+        Log.e(TAG, "Failed to write to file", e);
```
hayesjordan: Failed to close file

PR #836, opened 2016-09-20
Fix findbugs errors in collections library

cco3:
This also sets -e on the travis script, turning back on our
checks.


travis/build.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
```
hayesjordan: Looks like this might have caused a build error.

cco3:
Done...(travis can't do api 24 yet)

hayesjordan:
LGTM

PR #864, opened 2016-11-29
Set background image invisible when not in use

iankchristie:
Fixing bug that occurred when the device was turned. It would
show the background logo before redrawing the listview. This
sets the background textview to invisible after scanning is
completed.


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -101,6 +101,7 @@ public void run() {
         emptyGroupIdQueue();
         mSwipeRefreshWidget.setRefreshing(false);
```
cco3: Are these three methods always called together?  Could you just put them in their own method?

Issue #861, opened 2016-11-29
Custom PWS endpoint setting is confusing


cco3:
I think it'd make more sense if these settings only appeared when you have selected to use a custom PWS.

PR #869, opened 2016-12-14
Hide custom PWS settings when not in use

cco3:
When the user is not using a custom PWS, the custom PWS settings are
visible; confusing!  This change hides those settings when not in use.


cco3:
#861

PR #872, opened 2016-12-15
Uprade sdk

iankchristie:


android/PhysicalWeb/app/build.gradle
```diff
@@ -4,7 +4,7 @@ apply plugin: 'pmd'
 
 android {
     compileSdkVersion 25
-    buildToolsVersion "25.0.0"
+    buildToolsVersion "25.0.1"
```
cco3: 25.0.2

cco3: (I'm curious as to whether or not travis will build with this...we might have to change values until it passes)

PR #874, opened 2016-12-16
Fix Blocky UI behavior while scanning

iankchristie:
onUrlDeviceDiscoveryUpdate being called rapidly during scanning.
Because of inefficiencies in the process of adding it to the UI
this caused blocky behavior. This fix optimizes when we call the
UIThread to empty the GroupIdQueue.

### DIFF
--- a/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
+++ b/android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
@@ -296,32 +296,28 @@ public class NearbyBeaconsFragment extends ListFragment
 
   @Override
   public void onUrlDeviceDiscoveryUpdate() {
+    for (PwPair pwPair : mPwCollection.getGroupedPwPairsSortedByRank(
+        new Utils.PwPairRelevanceComparator())) {
+      String groupId = Utils.getGroupId(pwPair.getPwsResult());
+      Log.d(TAG, "groupid to add " + groupId);
+      if (mNearbyDeviceAdapter.containsGroupId(groupId)) {
+        mNearbyDeviceAdapter.updateItem(pwPair);
+      } else if (!mGroupIdQueue.contains(groupId)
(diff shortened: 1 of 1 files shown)


android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -296,32 +296,33 @@ public void onListItemClick(ListView l, View v, int position, long id) {
 
   @Override
   public void onUrlDeviceDiscoveryUpdate() {
+    Log.v(TAG, "onUrlDeviceDiscoveryUpdate");
+    if (SwipeDismissListViewTouchListener.isLocked()) {
```
cco3: I think this check needs to happen on the UI thread, and it's already handled by emptyGroupIdQueue, so just remove it.

android/PhysicalWeb/app/src/main/java/org/physical_web/physicalweb/NearbyBeaconsFragment.java
```diff
@@ -296,32 +296,33 @@ public void onListItemClick(ListView l, View v, int position, long id) {
...
+      Log.d(TAG, "groupid to add " + groupId);
+      if (mNearbyDeviceAdapter.containsGroupId(groupId)) {
+        mNearbyDeviceAdapter.updateItem(pwPair);
+      } else if (!mGroupIdQueue.contains(groupId)
+          && !Utils.isBlocked(pwPair)) {
+        mGroupIdQueue.add(groupId);
+      }
+    }
+    // If queue is empty or the second scan is not complete, do not
+    // try to empty queue. Causes blocky UI behavior
```
cco3: Remove the comment about blocky UI behavior...that's not the core reason to check, and it shouldn't inherently cause janky behavior.

iankchristie:
Is this fix ready to go?

cco3:
I think so.  Please squash the changes together first.

iankchristie:
@cco3 squashed the changes. Credit: Doug & Matt

PR #887, opened 2017-01-20
File serving

iankchristie:
allows to send large file over ble with multiple readcharacteristic requests.

### DIFF
--- a/nodejs/FatBeaconPeripheral/webpageCharacteristic.js
+++ b/nodejs/FatBeaconPeripheral/webpageCharacteristic.js
@@ -11,13 +11,24 @@ var WebpageCharacteristic = function() {
 
   this._value = Buffer.alloc(0);
   this._updateValueCallback = null;
+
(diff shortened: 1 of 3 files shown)


nodejs/FatBeaconPeripheral/webpageCharacteristic.js
```diff
@@ -11,13 +11,23 @@ var WebpageCharacteristic = function() {
 
   this._value = Buffer.alloc(0);
   this._updateValueCallback = null;
+  this._mtuSize = 251; //5 less than the specified mtu
```
cco3: space

cco3: (two before `//` and one after)

iankchristie:
@cco3 fixed changes. I also dismissed the review because it wasn't going away after I made the changes and then recommitted.

iankchristie:
don't know if that was the right move, but it happened.

PR #917, opened 2017-07-11
Fix broken links for blog posts

dvdhpkns:
Without http in front of the url clicking takes you to `https://github.com/google/physical-web/blob/master/blog.chromium.org/2015/07/exploring-physical-web-with-chrome-for.html`


dvdhpkns:
I signed it!

